PR #3, opened 2023-08-16
cp-1: Add web project starter

### LINKED ISSUE #1: Setup web project starter

liza-veis:


.eslintrc.yml
```diff
@@ -0,0 +1,131 @@
...
+    - blankLine: always
+      prev:
+        - const
+        - class
+      next: export
+  '@typescript-eslint/consistent-type-definitions':
+    - off
+  '@typescript-eslint/non-nullable-type-assertion-style':
+    - off
+  '@typescript-eslint/return-await':
```
what1s1ove, line 82: not sure if I need to add any additional rules at the beginning of the project. they are usually added as needed

backend/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "backend",
+  "private": true,
+  "type": "module",
+  "engines": {
+    "node": "18.x.x",
+    "npm": "9.x.x"
+  },
+  "imports": {
+    "#*": "./src/*"
```
what1s1ove, line 10: really bad choice. hash is a valid url parameter

import is a url

https://developer.mozilla.org/en-US/docs/Web/API/URL/hash

backend/src/libs/interfaces/entity.interface.ts
```diff
@@ -0,0 +1,6 @@
+interface Entity {
```
what1s1ove: if interfaces without any prefixes will be used, why use them then at all? what is better than types?

backend/package.json
v0ldemar01: https://betterprogramming.pub/the-native-way-to-configure-path-aliases-in-frontend-projects-5db70f19a6e0

v0ldemar01: https://abhijithota.me/posts/node-import-aliases/

.eslintrc.yml
v0ldemar01, line 1: Let's add this one
image

backend/src/libs/packages/config/libs/interfaces/config.interface.ts
```diff
@@ -0,0 +1,7 @@
+import { type Config as LibraryConfig } from 'shared/build/index.js';
+
+import { type EnvironmentSchema } from '../types/types.js';
+
+interface Config extends LibraryConfig<EnvironmentSchema> {}
```
v0ldemar01: Maybe type instead?

backend/src/packages/users/user.controller.ts
```diff
@@ -0,0 +1,65 @@
...
+  type APIHandlerResponse,
+  BaseController,
+} from '#libs/packages/controller/controller.js';
+import { HTTPCode } from '#libs/packages/http/http.js';
+import { type Logger } from '#libs/packages/logger/logger.js';
+import { type UserService } from '#packages/users/user.service.js';
+
+import { UsersApiPath } from './libs/enums/enums.js';
+
+/**
```
v0ldemar01, line 12: Can we move to another approach with swagger?

backend/src/packages/users/user.entity.ts
```diff
@@ -0,0 +1,88 @@
+import { type Entity } from '#libs/interfaces/interfaces.js';
+
+class UserEntity implements Entity {
+  private 'id': number | null;
```
v0ldemar01: let's use # directive for private fields of js

backend/src/packages/users/user.repository.ts
```diff
@@ -0,0 +1,47 @@
+import { type Repository } from '#libs/interfaces/interfaces.js';
+import { UserEntity } from '#packages/users/user.entity.js';
+import { type UserModel } from '#packages/users/user.model.js';
+
+class UserRepository implements Repository {
+  private userModel: typeof UserModel;
```
v0ldemar01, line 6: Same here

backend/tsconfig.json
```diff
@@ -0,0 +1,17 @@
...
+  "include": ["src", "knexfile.ts"],
+  "exclude": ["node_modules"],
+  "compilerOptions": {
+    "target": "ESNext",
+    "module": "ESNext",
+    "esModuleInterop": true,
+    "moduleResolution": "NodeNext",
+    "baseUrl": ".",
+    "paths": {
+      "#*": ["./src/*"]
```
v0ldemar01, line 12: Is it mandatory here?

backend/src/libs/packages/server-application/base-server-application-api.ts
```diff
@@ -0,0 +1,58 @@
...
+  private config: Config;
+
+  public constructor(
+    version: string,
+    config: Config,
+    ...handlers: ServerApplicationRouteParameters[]
+  ) {
+    this.version = version;
+    this.config = config;
+    this.routes = handlers.map((item) => ({
```
fshabanov:
```suggestion
    this.routes = handlers.map((handler) => ({
```
Maybe naming this variable something like 'handler' would be clearer
'item' is used in multiple files, same for others as well

backend/src/libs/packages/server-application/base-server-application.ts
```diff
@@ -0,0 +1,220 @@
...
+        return schema.validate(data, {
+          abortEarly: false,
+        });
+      };
+    });
+  }
+
+  private initErrorHandler(): void {
+    this.app.setErrorHandler(
+      (error: FastifyError | ValidationError, _request, replay) => {
```
fshabanov:
```suggestion
      (error: FastifyError | ValidationError, _request, reply) => {
```

.eslintrc.yml
what1s1ove: unicorn plugin has this rule by default

backend/package.json
what1s1ove: These are random articles from the Internet 😄
Node has added `node:` prefixes for the imports so that the imports become valid URLs. By adding the hash we kill what they added

backend/src/packages/users/user.entity.ts
```diff
@@ -0,0 +1,88 @@
...
+    id,
+    email,
+    passwordHash,
+    passwordSalt,
+  }: {
+    id: number;
+    email: string;
+    passwordHash: string;
+    passwordSalt: string;
+  }): UserEntity {
```
fshabanov, line 39: Maybe it's better to create separate types for this and other ones

frontend/index.html
```diff
@@ -0,0 +1,13 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="/vite.svg" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Vite + React + TS</title>
```
fshabanov:
```suggestion
    <title>CalmPal</title>
```

backend/package.json
liza-veis: Also Node has added native import aliases) How does this make a url invalid? Don't see the difference between this approach and usual aliases. If you mean that it's invalid because url has such a parameter, then I think they are very distinguishable, because # url parameter goes at the end

backend/src/libs/interfaces/entity.interface.ts
liza-veis: They are used for implementation, you can't implement types. And I also didn't get your point, how does prefix impact interface usage?

.eslintrc.yml
liza-veis: Yes, this can be resolved

backend/src/libs/packages/config/libs/interfaces/config.interface.ts
liza-veis: we use it to implement class from it, that's why

backend/src/packages/users/user.controller.ts
liza-veis: We have two options either leave it as it is or do it in a separate folder like it was on guruhub. The problem with the second solution is that it's much harder to maintain. I just chose the better one, to come up with another idea we need time for research here

backend/src/packages/users/user.entity.ts
liza-veis: I think we should not mix # with typescript access modifiers, and either use javascript approach or typescript

backend/package.json
what1s1ove: I would say that we need to avoid any characters that can be used in the URL, since the Node tries to do everything so that the path in the import will be a valid URL. We make it worse

backend/tsconfig.json
liza-veis: it's not required, this is done for imports suggestions, there is an open issue for it, this is a quick solution before it's resolved

backend/src/libs/interfaces/entity.interface.ts
what1s1ove: I can implement type 🤔

image

> how does prefix impact interface usage?

No, it doesn't. But right now in the application, there is no difference between interface and type. Why do we need to use interfaces then?

backend/package.json
liza-veis: If they try, then why did they add this feature?

liza-veis: it's not an experimental technology

what1s1ove: > If they try, then why did they add this feature?

it was added before the `node:` protocol. Everyone can make mistakes 🥲

backend/src/libs/interfaces/entity.interface.ts
liza-veis: Ok, then it's more semantically correct to use interface because its goal is to define a contract and also it restricts you to a plain object, while type can have other options that won't work with implements. So for me, it looks like type defines either data type or structure, while interface definine a contract that can be used only for implementation

backend/package.json
liza-veis: not sure if this was a mistake

what1s1ove: they want to make the import path a valid URL. Of course, the hash in the import is a mistake 😄

backend/src/libs/interfaces/entity.interface.ts
what1s1ove: Okay, `import { User, Admin, Doctor, Man } from 'packages/users'`, in this case, where is the interface? I do not know 🤔

backend/package.json
liza-veis: how it defers from other import aliases?

.eslintrc.yml
liza-veis: yes, you are right, will change this

backend/src/libs/interfaces/entity.interface.ts
liza-veis: that's a good point, will change them to types

.eslintrc.yml
liza-veis: This is needed from the start, it requires to use ! instead of casting, or possible null types
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/non-nullable-type-assertion-style.md

backend/src/libs/interfaces/entity.interface.ts
liza-veis: Changed this

backend/src/libs/packages/config/libs/interfaces/config.interface.ts
liza-veis: UPD: Replaced with types

backend/src/libs/packages/server-application/base-server-application-api.ts
liza-veis: Changed

backend/src/libs/packages/server-application/base-server-application.ts
liza-veis: Changed

frontend/index.html
liza-veis: Changed

backend/src/packages/users/user.entity.ts
liza-veis: I think it is not needed here, we will only use them once and we will need to define a lot of such types, at least 5 (but can be much more). So it will be hard to navigate this file

backend/src/packages/users/user.repository.ts
liza-veis: answered above

backend/src/packages/users/user.entity.ts
v0ldemar01: So which one?)

liza-veis: typescript because it has other modifiers, they are more obvious and we can lint them. But if most of us prefer js, I don't mind

backend/package.json
liza-veis: I would prefer to leave them because:
- they work faster
- work without build in node environment
- no need for additional libraries
- imports autocomplete much better (the only thing is that suggestions don't work, but this easily can be fixed with ts paths, until the issue is resolved)

backend/src/packages/users/user.entity.ts
liza-veis: Can be resolved, we decided to leave typescript approach

backend/src/packages/users/user.repository.ts
liza-veis: can be resolved too

v0ldemar01 (approved):

fshabanov (approved):

PR #6, opened 2023-08-17
cp-4: Setup deployment for development environment

v0ldemar01:


.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  ci:
+  linter:
```
liza-veis: Why linter 🥲 ? we usually name the same as file name, I think it's better to leave ci

.github/workflows/cd-development.yml
```diff
@@ -0,0 +1,66 @@
+name: Continuous Delivery
```
liza-veis: We either should remove development from file name or add it to name and job name

docker-compose.services.yml
```diff
@@ -0,0 +1,25 @@
...
+services:
+  calmpal_db:
+    container_name: calmpal_db
+    image: postgres:15.2
+    environment:
+      - POSTGRES_DB=$(echo "$DB_CONNECTION_STRING" | awk -F[@/:] '{print $8}')
+      - POSTGRES_USER=$(echo "$DB_CONNECTION_STRING" | awk -F[@/:] '{print $4}')
+      - POSTGRES_PASSWORD=$(echo "$DB_CONNECTION_STRING" | awk -F[@/:] '{print $5}')
+    ports:
+      - 5434:5432
```
liza-veis, line 12: Will this work if we have a different connection string in .env?

liza-veis (changes requested):

.github/workflows/ci.yml
v0ldemar01: So that file contains 2 jobs that are related to ci

v0ldemar01: I can split into 2 ones: ci and ci-build?

docker-compose.services.yml
v0ldemar01: Good point) Will change into .env.example

v0ldemar01: I had postgres locally with 5432 port, so to prevent ambiguous port mapping

.github/workflows/ci.yml
liza-veis: ok, got it, let's split them not to run it on every trigger

docker-compose.yml
```diff
@@ -1,11 +0,0 @@
-version: '3'
```
fshabanov, line 1: I guess this line from readme should change, since this file is now deleted:
```docker-compose -f docker-compose.yml up --build```

File: backend/.env.example [hunk 5]
```diff
@@ -1,14 +1,21 @@
...
-NODE_ENV=development
+NODE_ENV=local
 PORT=3001
 HOST=localhost
 
 #
 # DATABASE
 #
-DB_CONNECTION_STRING=[db_client]://[db_username]:[db_user_password]@localhost:[db_port]/[db_name]
+DB_USERNAME=db_username]
```
liza-veis:
```suggestion
DB_USERNAME=db_username
```

backend/src/libs/packages/server-application/libs/constants/api.constants.ts
```diff
@@ -13,7 +11,7 @@ const WHITE_ROUTES = [
     methods: ['POST'],
   },
   { path: new RegExp(`^/api/v1${APIPath.DOCUMENTATION}/*`), methods: ['GET'] },
-  { path: new RegExp(SERVED_PAGE_PATH), methods: ['GET'] },
+  { path: new RegExp('/health$'), methods: ['GET'] },
```
liza-veis:
```suggestion
  { path: new RegExp('^/health$'), methods: ['GET'] },
```

liza-veis (changes requested):

backend/.env.example
```diff
@@ -1,14 +1,21 @@
...
 #
-DB_CONNECTION_STRING=[db_client]://[db_username]:[db_user_password]@localhost:[db_port]/[db_name]
+DB_USERNAME=db_username
+DB_PASSWORD=db_user_password
+DB_HOST=localhost
+DB_PORT=db_port
+DB_NAME=db_name
+DB_DIALECT=pg
+DB_POOL_MIN=2
+DB_POOL_MAX=10
```
fshabanov: These fields are repeated twice

liza-veis (approved):

fshabanov (approved):

PR #8, opened 2023-08-18
cp-2: Fix mobile imports

liza-veis:


shared/tsconfig.json
```diff
@@ -4,7 +4,7 @@
   "exclude": ["node_modules"],
   "compilerOptions": {
     "target": "ESNext",
-    "moduleResolution": "NodeNext",
+    "moduleResolution": "Node16",
```
v0ldemar01, line 7: With NodeNext it is not working properly?

liza-veis: Yes, eslint will automatically consider this package as esm. So we can either remove type module, or change module resolution

what1s1ove (approved):

fshabanov (approved):

PR #31, opened 2023-08-19
cp-20: Add Loading Component

ossavemu:
The Loading component was created to show a complete screen with an animated component with css in a loop. The loader.tsx and loader.scss files were created. Added loader to the style index. 

Captura de pantalla 2023-08-19 145033
Captura de pantalla 2023-08-19 145106

PR #33, opened 2023-08-19
cp-32: + added Header

ossavemu:
Captura de pantalla 2023-08-19 165750

Header with navigation bar created. An attempt has been made to follow the styles of the example. For the screenshot, the background of the screen was colored in an auxiliary way, but the white background was kept.


frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,22 @@
+import '#assets/css/header.scss';
+
+import reactLogo from '#assets/img/react.svg';
```
liza-veis: Replace with the logo attached to the issue

frontend/src/assets/css/header.scss
```diff
@@ -0,0 +1,31 @@
+$navbar-bg: #f8f9fa;
```
liza-veis: use css variables, also move common colors to global variables as I described here https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/31#discussion_r1299275782

frontend/src/assets/css/header.scss
```diff
@@ -0,0 +1,31 @@
...
+.navbar-content {
+  display: flex;
+  justify-content: space-between;
+  padding-right: 10%;
+  padding-left: 10px;
+}
+
+.nav-link {
+  padding: 5px 10px;
+  color: dimgray;
```
liza-veis: use color from design

frontend/src/assets/css/header.scss
```diff
@@ -0,0 +1,31 @@
...
+  text-decoration: none;
+  border-radius: 5px;
+  transition: background-color 0.3s ease;
+}
+
+.nav-link:hover {
+  color: black;
```
liza-veis: move to global css variables

frontend/src/assets/css/header.scss
```diff
@@ -0,0 +1,31 @@
...
+  background-color: $navbar-hover-bg;
```
liza-veis: we will have these styles for link by default, so move them to Link component  (except padding)

frontend/src/assets/css/header.scss
liza-veis: header.module.scss should be in header component folder, we will use scss modules for components

frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { AppRoute } from '#libs/enums/app-route.enum.js';
+
+import { Link } from '../components.js';
+
+const Header: React.FC = () => {
+  return (
+    <header className="navbar">
+      <nav className="navbar-content">
```
liza-veis:
```suggestion
    <header className="header">
      <nav className="nav">
```

frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,22 @@
...
+        <img src={reactLogo} className="App-logo" width="30" alt="logo" />
```
liza-veis: Let's move logo outside nav, also logo should be wrapped into link to the root

liza-veis (changes requested):

frontend/src/assets/css/header.scss
ossavemu: Waiting for response on the other pull request

frontend/src/assets/css/header.scss
ossavemu: I use this tool to pick the color: 
Captura de pantalla 2023-08-19 222355

frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+const Header: React.FC = () => {
+  return (
+    <header className={styles.header}>
+      <Link to={AppRoute.ROOT}>
+        <img src={headerLogo} width="40" alt="logo" />
+      </Link>
+      <nav className={styles.nav}>
+        <Link className={styles.link} to={AppRoute.SIGN_IN}>
+          Sing In
```
fshabanov:
```suggestion
          Sign In
```

frontend/src/libs/components/link/link.tsx
```diff
@@ -3,13 +3,24 @@ import { NavLink } from 'react-router-dom';
 import { type AppRoute } from '#libs/enums/enums.js';
 import { type ValueOf } from '#libs/types/types.js';
 
+import styles from './link.module.scss';
```
fshabanov:
```suggestion
import styles from './styles.module.scss';
```

frontend/src/libs/components/link/link.tsx
```diff
@@ -3,13 +3,24 @@ import { NavLink } from 'react-router-dom';
...
+
 type Properties = {
   to: ValueOf<typeof AppRoute>;
   children: React.ReactNode;
+  className?: string;
 };
 
-const Link: React.FC<Properties> = ({ children, to }: Properties) => (
-  <NavLink to={to}>{children}</NavLink>
+const stylePatcher = (newStyle: string): string => `${styles.link} ${newStyle}`;
```
fshabanov: This function can be used in other components as well. Move it to helpers folder

frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,22 @@
+import headerLogo from '#assets/img/header-logo.svg';
+import { AppRoute } from '#libs/enums/app-route.enum.js';
+
+import { Link } from '../components.js';
+import styles from './header.module.scss';
```
fshabanov:
```suggestion
import styles from './styles.module.scss';
```

fshabanov (changes requested):

frontend/src/libs/components/link/link.tsx
```diff
@@ -1,15 +1,30 @@
...
-const Link: React.FC<Properties> = ({ children, to }: Properties) => (
-  <NavLink to={to}>{children}</NavLink>
-);
+const Link: React.FC<Properties> = ({
+  children,
+  to,
+  className = '',
+}: Properties) => {
+  const { link } = styles;
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
```
v0ldemar01: You shouldn't do in such way

frontend/src/libs/components/link/styles.module.scss
```diff
@@ -0,0 +1,8 @@
+@import url("../../../assets/css/variables.scss");
```
v0ldemar01: Single quotes?

v0ldemar01 (changes requested):

frontend/src/libs/components/link/link.tsx
```diff
@@ -1,15 +1,30 @@
...
+  const mergedStyles = mergeStyles(className ?? '', link);
```
v0ldemar01: Maybe className from props should to override the link one?

v0ldemar01 (changes requested):

v0ldemar01 (changes requested):
Are you sure that you have ran precommit hook script?

frontend/src/libs/components/link/link.tsx
ossavemu: yes it does. In helper function, the new styles are join to the end, so there is override. The idea is not to overwrite completely, but the repeated properties, that's why I put them at the end

ossavemu:
Yes, I am using everything well. But it seems that the CI rules in github (after doing the push) are different from the local project.

frontend/src/libs/components/link/styles.module.scss
ossavemu: fixed

ossavemu:
New screenshots:
Captura de pantalla 2023-08-20 094314

frontend/src/assets/css/variables.scss
```diff
@@ -0,0 +1,7 @@
+:root {
+  --navbar-bg: #f8f9fa;
+  --navbar-shadow: 0 2px 4px #0000001a;
+  --navbar-hover-bg: #e2e3e5;
+  --hover-color: #000000;
+  --link-text-color: #7c7f89;
```
fshabanov: The same colors can be used in different places. Do not include navbar, link in the name.

frontend/src/libs/components/header/styles.module.scss
```diff
@@ -0,0 +1,27 @@
+@import url("../../../assets/css/variables.scss");
```
fshabanov: No need to include this file. It should be imported in global styles.scss file, which is already imported to the app.

frontend/src/libs/helpers/merge-styles.ts
```diff
@@ -0,0 +1,6 @@
+type Style = string | undefined;
+
+const mergeStyles = (newStyle: Style, defaultStyle: Style): string =>
```
fshabanov:
```suggestion
const mergeStyles = (newStyle?: string, defaultStyle?: string): string =>
```

fshabanov (changes requested):

frontend/src/libs/helpers/merge-styles.ts
ossavemu: All implemented

frontend/src/assets/css/header.scss
liza-veis: it's better to use figma colors, not side color picker. To do this you need to select the element and you will see it's color in the side panel
Screenshot 2023-08-20 at 19 21 56

frontend/src/assets/css/variables.scss
```diff
@@ -0,0 +1,7 @@
+:root {
+  --white-pale: #f8f9fa;
```
liza-veis: It's totally white on design 
```suggestion
  --white: #ffffff;
```
Screenshot 2023-08-20 at 19 28 13

frontend/src/assets/css/variables.scss
```diff
@@ -0,0 +1,7 @@
...
+  --black-shadow: 0 2px 4px #0000001a;
```
liza-veis:
```suggestion
  --shadow: 0 2px 4px #0000001a;
```

frontend/src/assets/css/variables.scss
```diff
@@ -0,0 +1,7 @@
...
+  --lightgray: #e2e3e5;
+  --black: #000000;
+  --medium-gray: #7c7f89;
```
liza-veis: We have a convention that we name color variables like gray-100, where number is intensity of the color (we chose ourselves), except white and black colors
```suggestion
  --black: #000000;
  --gray-100: #e2e3e5;
  --gray-200: #7c7f89;
```

frontend/src/assets/img/header-logo.svg
liza-veis, line 1: We may use it not only in header, so please rename the file to logo.svg

frontend/src/libs/components/header/styles.module.scss
```diff
@@ -0,0 +1,25 @@
...
+
+.link {
+  margin: 5px 0;
+  padding: 5px 10px;
+}
+
+.link:hover {
+  color: var(--black);
+  background-color: var(--lightgray);
+}
```
liza-veis: padding and link:hover styles should be in link component

frontend/src/libs/helpers/merge-styles.ts
```diff
@@ -0,0 +1,4 @@
+const mergeStyles = (newStyle?: string, defaultStyle?: string): string =>
```
liza-veis: Let's use clsx lib for this, also please rename helper to getValidClassNames

frontend/src/libs/components/link/styles.module.scss
```diff
@@ -0,0 +1,8 @@
+.link {
```
liza-veis, line 1: We have some font properties that we need to set too, to make it look like on design. Pay attention to the text block when you select Sign In in figma. So you also need to:

1. setup custom fonts
 - download Monserat fonts and add them to public/fonts folder
 - create fonts.scss in assets/css/ and define via font-face
 2. add font, font-size, font-weight, line-height properties to the link styles with values from figma design

Screenshot 2023-08-20 at 19 39 20

liza-veis (changes requested):

frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,23 @@
+import headerLogo from '#assets/img/header-logo.svg';
+import { AppRoute } from '#libs/enums/app-route.enum.js';
```
fshabanov:
```suggestion
import { AppRoute } from '#libs/enums/enums.js';
```

fshabanov (changes requested):

frontend/src/libs/components/link/styles.module.scss
ossavemu: The official docs have font-face definitions: https://github.com/JulietaUla/Montserrat/tree/master/fonts/webfonts I use Montserrat.css

ossavemu: I modified the css to keep only the normal fonts with woff2 format (the lightest one).

frontend/src/libs/components/header/header.tsx
ossavemu: solved

frontend/src/libs/components/link/styles.module.scss
ossavemu: solved

frontend/src/assets/css/fonts.scss
```diff
@@ -0,0 +1,143 @@
+@font-face {
```
liza-veis, line 1: Let's leave only 600/400 without italic as we have on our figma design, so that we don't load fonts that we don't use

frontend/src/libs/components/link/styles.module.scss
```diff
@@ -0,0 +1,18 @@
...
+  display: flex;
+  align-items: center;
+  padding: 5px 10 px;
```
liza-veis: I don't see styles for link state in figma, so let's keep it simple for now: remove padding and background-color changing on hover

frontend/src/libs/components/link/styles.module.scss
```diff
@@ -0,0 +1,18 @@
+.link {
+  display: flex;
+  align-items: center;
```
liza-veis: we don't need display flex for the link
```suggestion
```

frontend/src/libs/components/link/link.tsx
```diff
@@ -1,15 +1,29 @@
...
 import { type AppRoute } from '#libs/enums/enums.js';
+import { getValidClassNames } from '#libs/helpers/helpers.js';
 import { type ValueOf } from '#libs/types/types.js';
 
+import styles from './styles.module.scss';
+
 type Properties = {
   to: ValueOf<typeof AppRoute>;
   children: React.ReactNode;
+  className?: string | undefined;
```
fshabanov:
```suggestion
  className?: string;
```
'?' already means it can be 'undefined'

frontend/src/libs/components/link/link.tsx
```diff
@@ -1,15 +1,29 @@
...
 };
 
-const Link: React.FC<Properties> = ({ children, to }: Properties) => (
-  <NavLink to={to}>{children}</NavLink>
-);
+const Link: React.FC<Properties> = ({
+  children,
+  to,
+  className = '',
+}: Properties) => {
```
fshabanov:
```suggestion
}) => {
```
FC<Properties> already uses the types for the props

frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,23 @@
+import headerLogo from '#assets/img/logo.svg';
```
liza-veis:
```suggestion
import logo from '#assets/img/logo.svg';
```

frontend/src/libs/helpers/get-valid-class-names.ts
```diff
@@ -0,0 +1,6 @@
+import clsx from 'clsx';
+
+const getValidClassNames = (newStyle?: string, defaultStyle?: string): string =>
+  clsx(defaultStyle, newStyle);
```
fshabanov: What if we have more than 2 styles? Consider this possibility as well

frontend/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { AppRoute } from '#libs/enums/enums.js';
+
+import { Link } from '../components.js';
```
liza-veis:
```suggestion
import { Link } from '#libs/components/components.js';
```

fshabanov (changes requested):

fshabanov:
Please, fix the merge conflicts as well

frontend/src/libs/helpers/get-valid-class-names.ts
Liza-Veis, line 4:
```suggestion
const getValidClassNames = (...inputs: ClassValue[]): string =>
  clsx(...inputs);
```

frontend/src/libs/components/link/link.tsx
```diff
@@ -1,15 +1,29 @@
...
+  const { link } = styles;
+  const mergedStyles = getValidClassNames(className, link);
```
liza-veis:
```suggestion
  const className = getValidClassNames(className, styles.link);
```

liza-veis (changes requested):

frontend/public/fonts/montserrat/Montserrat-Black.woff2
liza-veis, line 1: All files should be in kebab-case - 'Montserrat-Black' -> 'montserrat-black'

liza-veis (changes requested):

frontend/src/libs/components/link/link.tsx
ossavemu: yes, but i had this lint error: 'className' is missing in props validationeslint[react/prop-types](https://github.com/jsx-eslint/eslint-plugin-react/tree/master/docs/rules/prop-types.md)
(parameter) className: string

ossavemu: image

frontend/src/libs/helpers/get-valid-class-names.ts
liza-veis: Also we may want to use not only strings but also conditions, so please use something like 
```
const getValidClassNames = (...inputs: ClassValue[]): string => clsx(...inputs);
```

frontend/src/libs/components/link/styles.module.scss
ossavemu: I don't have the link to the figma yet

frontend/src/libs/components/link/link.tsx
liza-veis: @ossavemu  disable this rule in frontend eslintrc (FYI you can change configs only when one of the coaches ask you)

ossavemu: done :)

frontend/src/libs/components/link/styles.module.scss
liza-veis: in the slack calmpal_team channel we have all links pinned to the top. Please let me know if you still aren't able to get figma link
Screenshot 2023-08-21 at 00 21 54

frontend/src/libs/components/link/link.tsx
ossavemu: repeated names

frontend/src/libs/components/link/link.tsx
ossavemu: The problem is with the eslintrc.yml, for some reason when modifying it this error appears. I have tried many ways but I can't figure out how to commit after editing it.
I then restated the type in the FC link props.

ossavemu:
image


That is the reason why I use destructuring. If I follow the recommendations of that lint from the CI github, the local project lintern prevents me from doing so. Destructuring passed both tests.

frontend/src/libs/components/link/link.tsx
Liza-Veis: You should add your updated eslintrc to git stage

liza-veis:
@ossavemu use styles['style'] instead of destructuring

frontend/src/libs/components/link/link.tsx
liza-veis: Don't forget to stage your files

v0ldemar01:
> @ossavemu use styles['style'] instead of destructuring

We have disabled the rule of dot notation eslint rule

frontend/src/libs/helpers/get-valid-class-names.ts
```diff
@@ -0,0 +1,6 @@
+import { type ClassValue } from 'clsx';
+import clsx from 'clsx';
```
v0ldemar01:
```suggestion
import { type ClassValue, clsx } from 'clsx';
```
Can we use it in such way?

v0ldemar01 (changes requested):

frontend/src/libs/helpers/get-valid-class-names.ts
ossavemu: Done!

frontend/src/libs/components/link/link.tsx
ossavemu: I did a test where I passed classes to the Link component and it showed me a lint error that is removed when undefined is explicitly defined in the type.

Here is the error: 

The type '{ children: element; to:"/"; className: string | undefined; }' cannot be assigned to the 'Properties' type with 'exactOptionalPropertyTypes: true'. Consider adding 'undefined' to the target property types.

The way to fix it would be every time you import a style do it like this: 

`<Link to={AppRoute.ROOT} className={styles['image'] ?? ''}>`

But I don't know, I think it looks a bit weird. I don't know if it's better to declare undefined explicitly.

frontend/src/libs/components/link/link.tsx
ossavemu: Solved!

frontend/src/libs/components/link/link.tsx
fshabanov: className?: string; is exactly the same as className?: string | undefined;
? means ```type | undefined```
So there's no need to mention 'undefined' once again there

ossavemu: Yes, but when you test this ability, the error I mentioned appears.

Could be related with: 'exactOptionalPropertyTypes: true'

this is another way to solve: `<Link to={AppRoute.SIGN_IN} className={styles['image'] as string}>`

liza-veis (dismissed):
Nice work) 🙌

v0ldemar01 (dismissed):

fshabanov (dismissed):

fshabanov (approved):

liza-veis (approved):

frontend/src/assets/css/header.scss
dmytro-kucherenko: @Liza-Veis Do we really have these styles for links by default? At least in sign up/sign in redirecting link it should be white and it doesn't seem to become black on hovering.

PR #34, opened 2023-08-20
cp-26: Add Loader

Vika030718:
- The enum for global colors was created
- The Loader Component was created
Screenshot 2023-08-20 at 13 38 02

### DIFF
--- a/mobile/src/libs/components/loader/loader.tsx
+++ b/mobile/src/libs/components/loader/loader.tsx
@@ -1,10 +1,10 @@
 import React from 'react';
 import { ActivityIndicator } from 'react-native';
 
-import { AppColors } from '#libs/enums/ui/app-colors';
+import { AppColor } from '#libs/enums/enums';
 
 const Loader: React.FC = () => {
-  return <ActivityIndicator size="large" color={AppColors.BLUE_100} />;
+  return <ActivityIndicator size="large" color={AppColor.BLUE_100} />;
 };
 
 export { Loader };
--- a/mobile/src/libs/components/loader/loader.tsx
+++ b/mobile/src/libs/components/loader/loader.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { ActivityIndicator, Text, View } from 'react-native';
+import { ActivityIndicator, View } from 'react-native';
 
 import { AppColor } from '#libs/enums/enums';
 
@@ -8,9 +8,7 @@ import { styles } from './styles';
 const Loader: React.FC = () => {
   return (
     <View style={styles.container}>
-      <Text>
-        <ActivityIndicator size="large" color={AppColor.BLUE_100} />;
-      </Text>
+      <ActivityIndicator size="large" color={AppColor.BLUE_100} />
     </View>
   );
 };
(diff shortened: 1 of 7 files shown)


liza-veis:
Please attach a screenshot

mobile/src/libs/enums/ui/app-colors.ts
```diff
@@ -0,0 +1,12 @@
+const AppColors = {
```
liza-veis: enum name should be in sing form - AppColor, also rename file to app-color.enum.ts

mobile/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { ActivityIndicator } from 'react-native';
+
+import { AppColors } from '#libs/enums/ui/app-colors';
```
liza-veis: you should import AppColor from '#libs/enums/enums'

liza-veis (changes requested):

mobile/src/libs/enums/ui/app-colors.ts
fshabanov: Create a ui.ts file and export AppColor from that file

fshabanov (changes requested):

liza-veis:
You should fix ci errors, you can know what's needed to be fixed when you click on failed workflow details
Screenshot 2023-08-20 at 21 19 12

v0ldemar01 (dismissed):

fshabanov:
Usually we will not display other components while displaying the loader. Could you confirm that we will see the loader in the center of the screen if this is the case? From what I see I'm not sure about it

mobile/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { ActivityIndicator } from 'react-native';
+
+import { AppColor } from '#libs/enums/enums';
+
+const Loader: React.FC = () => {
+  return <ActivityIndicator size="large" color={AppColor.BLUE_100} />;
```
liza-veis: As @fshabanov mentioned, please wrap it into View and add styles so it will take full component size and will show loader itself at the center

liza-veis (changes requested):

mobile/src/libs/components/loader/loader.tsx
Vika030718: Done, thank you.
Screenshot 2023-08-20 at 14 33 58

mobile/src/libs/components/loader/loader.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { ActivityIndicator, Text, View } from 'react-native';
+
+import { AppColor } from '#libs/enums/enums';
+
+import { styles } from './styles';
+
+const Loader: React.FC = () => {
+  return (
+    <View style={styles.container}>
+      <Text>
```
liza-veis: remove Text component
```suggestion
```

liza-veis (changes requested):

fshabanov (approved):

liza-veis (approved):

v0ldemar01 (approved):

what1s1ove (approved):

PR #35, opened 2023-08-20
cp-21: Implementing error handling

juanc-escobar:
**Changes:**

- Installed the **react-toastify** package.
- Integrated **ToastContainer** in the index file.
- Implemented **error-handling-middleware.ts.**
- Modified the Redux store to incorporate the error handling middleware.
- Styled toast notifications to align with the Figma designs.

**Challenges & Observations:**

- Ran into a lot of linter errors that kept popping up as I made changes. It took quite a bit of time to sort them all out.
- At first, I wasn't too clear on what the task was fully about, which made me unsure about how to go about it. This added some extra time to the process.

**Visuals:**

cp-21


package.json
```diff
@@ -49,5 +49,8 @@
   "simple-git-hooks": {
     "pre-commit": "npx lint-staged && npm run lint:type",
     "commit-msg": "npx commitlint --edit $1"
+  },
+  "dependencies": {
+    "react-toastify": "9.1.3"
```
fshabanov: This package should be added to the frontend folder

frontend/src/middlewares/error-handling/error-handling-middleware.ts
```diff
@@ -0,0 +1,22 @@
...
+  string,
+  unknown,
+  Error | undefined
+>;
+
+const handleAction =
+  (next: (action: RejectedAction) => RejectedAction) =>
+  (action: RejectedAction): RejectedAction => {
+    if (action.type.endsWith('/rejected') && action.error) {
+      toast.error(action.error.message);
```
fshabanov: Let's create a service for notifications. You can check how it was implemented in the mini project

frontend/src/assets/css/scaffolding.scss
```diff
@@ -14,6 +14,10 @@ body {
 }
 
 #root {
+  --toastify-color-error: #d51a52;
+  --toastify-icon-color-error: var(--toastify-color-error);
+  --toastify-color-progress-error: var(--toastify-color-error);
```
fshabanov: Global css variables should be assigned in assets/css/variables.scss file

frontend/src/index.tsx
```diff
@@ -37,6 +39,7 @@ createRoot(document.querySelector('#root') as HTMLElement).render(
           },
         ]}
       />
+      <ToastContainer />
```
fshabanov: Let's create a separate component for the toaster and use ToastContainer in that component

fshabanov (changes requested):

fshabanov:
You should assign the PR to yourself in future

frontend/src/middlewares/error-handling/error-handling-middleware.ts
```diff
@@ -0,0 +1,22 @@
...
+    }
+    return next(action);
+  };
+
+const errorHandlingMiddleware: Middleware = () => handleAction;
```
v0ldemar01:
```suggestion
const handleError: Middleware = () => handleAction;
```

frontend/src/middlewares/error-handling/error-handling-middleware.ts
```diff
@@ -0,0 +1,22 @@
...
+  unknown,
+  string,
+  unknown,
+  Error | undefined
+>;
+
+const handleAction =
+  (next: (action: RejectedAction) => RejectedAction) =>
+  (action: RejectedAction): RejectedAction => {
+    if (action.type.endsWith('/rejected') && action.error) {
```
v0ldemar01: Create an action which contains an notification service to notify via toast

v0ldemar01: Please use 1 function instead of 2 ones)

v0ldemar01: Please prefer curly and return to increase readness

v0ldemar01 (changes requested):

frontend/src/libs/packages/store/store.package.ts
```diff
@@ -38,13 +39,14 @@ class Store {
         auth: authReducer,
         users: usersReducer,
       },
-      middleware: (getDefaultMiddleware) => {
-        return getDefaultMiddleware({
+      middleware: (getDefaultMiddleware) => [
+        ...getDefaultMiddleware({
```
HIMDanny, line 43: as far as I know, using the spread operator to add middleware can cause problems with dispatch function being typed correctly. Maybe it is better to use concat instead of spread operator? Therefore, there will be no problem with inferring the dispatch function in the future.

Moreover, it is a suggested way to define middleware from redux toolkit docs: https://redux-toolkit.js.org/usage/usage-with-typescript/#correct-typings-for-the-dispatch-type

package.json
liza-veis: You can do this by passing workspace to the command - `npm run install <dep name> -w <workspace-name> `

frontend/src/middlewares/error-handling/error-handling-middleware.ts
```diff
@@ -0,0 +1,22 @@
+import { type Middleware, type PayloadAction } from '@reduxjs/toolkit';
```
liza-veis: Move file to frontend/src/libs/packages/store/middlewares/handle-error.middleware.ts

frontend/src/middlewares/error-handling/error-handling-middleware.ts
```diff
@@ -0,0 +1,22 @@
...
+
+type RejectedAction = PayloadAction<
+  unknown,
+  string,
+  unknown,
+  Error | undefined
+>;
+
+const handleAction =
+  (next: (action: RejectedAction) => RejectedAction) =>
```
liza-veis:
```suggestion
  (next: typeof store.instance.dispatch =>
```

frontend/src/middlewares/error-handling/error-handling-middleware.ts
```diff
@@ -0,0 +1,22 @@
...
+  (action: RejectedAction): RejectedAction => {
```
liza-veis: Use AnyAction from redux-toolkit instead of rejected action
```suggestion
  (action: AnyAction): unknown => {
```

frontend/src/middlewares/error-handling/error-handling-middleware.ts
liza-veis: Let's use such a condition, redux toolkit already has helpers to work with actions

```suggestion
    if (isRejected(result) && action.error) {
```

liza-veis (changes requested):

Liza-Veis:
Also FYI the error from the attached screenshot means that your backend wasn't connect to the database

Vika030718 (dismissed):

frontend/src/libs/packages/store/middlewares/handle-error.middleware.ts
```diff
@@ -0,0 +1,22 @@
...
+} from '@reduxjs/toolkit';
+import { isRejected } from '@reduxjs/toolkit';
+
+import { NotificationService } from '#libs/helpers/notification-service.js';
+
+const handleAction =
+  (next: Dispatch): ((action: AnyAction) => AnyAction) =>
+  (action: AnyAction): AnyAction => {
+    if (isRejected(action) && action.error.message) {
+      NotificationService.error(action.error.message);
```
v0ldemar01: Use dispatch to call action that contains NotificationService.error(action.error.message);

frontend/src/libs/packages/store/middlewares/handle-error.middleware.ts
```diff
@@ -0,0 +1,22 @@
...
+      return next(action);
+    }
+    return next(action);
+  };
+
+const handleError: Middleware = () => handleAction;
```
v0ldemar01: Let's proceed with only handleError without handleAction

v0ldemar01: Get dispatch form first callback param, add a type for it like AppDispatch(create this common type AppDispatch - typeof store.dispatch)

frontend/src/libs/packages/store/middlewares/handle-error.middleware.ts
```diff
@@ -0,0 +1,22 @@
...
+  type AnyAction,
+  type Dispatch,
+  type Middleware,
+} from '@reduxjs/toolkit';
+import { isRejected } from '@reduxjs/toolkit';
+
+import { NotificationService } from '#libs/helpers/notification-service.js';
+
+const handleAction =
+  (next: Dispatch): ((action: AnyAction) => AnyAction) =>
```
v0ldemar01: Use curly and return for declaration

v0ldemar01 (changes requested):

frontend/src/index.tsx
```diff
@@ -1,3 +1,4 @@
+import 'react-toastify/dist/ReactToastify.css';
```
fshabanov: You can move this import to Toaster component

frontend/src/libs/helpers/notification-service.ts
```diff
@@ -0,0 +1,9 @@
+import { toast } from 'react-toastify';
+
+const NotificationService = {
+  error(message: string): void {
+    toast.error(message);
+  },
+};
```
fshabanov: Create it as a class inside #libs/packages folder, not as a helper function

frontend/src/libs/packages/store/store.package.ts
```diff
@@ -7,6 +7,7 @@ import { configureStore } from '@reduxjs/toolkit';
 
 import { AppEnvironment } from '#libs/enums/enums.js';
 import { type Config } from '#libs/packages/config/config.js';
+import { handleError } from '#libs/packages/store/middlewares/handle-error.middleware.js';
```
fshabanov:
```suggestion
import { handleError } from '#libs/packages/store/middlewares/middlewares.js';
```

fshabanov (changes requested):

frontend/src/assets/css/variables.scss
```diff
@@ -1,4 +1,7 @@
 :root {
+  --toastify-color-error: #d51a52;
```
dmytro-kucherenko: We already have this hex as default error color, it's called --red-300

frontend/src/assets/css/scaffolding.scss
```diff
@@ -31,3 +31,13 @@ body {
   clip: rect(0 0 0 0);
   clip-path: inset(100%);
 }
+
+.Toastify__toast--error {
+  color: #222222;
```
liza-veis: Let's use color from our design, we already have variables in assets/variables.scss

```suggestion
  color: #222222;
```

frontend/src/assets/css/variables.scss
liza-veis: As Dmytro said, we already have this color, and let's not complicate the logic with variables, so just remove all these three toastify variables and use --red-300

frontend/src/assets/css/scaffolding.scss
```diff
@@ -31,3 +31,13 @@ body {
...
   clip-path: inset(100%);
 }
+
+#root .Toastify__toast--error {
+  color: #222222;
+  font-weight: 400;
+  font-size: 12px;
+  font-family: Montserrat, sans-serif;
+  font-style: normal;
+  line-height: normal;
```
liza-veis:
```suggestion
  line-height: 1.5; // or other your value
```

frontend/src/assets/css/variables.scss
```diff
@@ -11,4 +11,7 @@
   --pink-200: #e1a7d5;
   --red-300: #d51a52;
   --white-alpha-80: #ffffff80;
+  --toastify-color-error: var(--red-300);
+  --toastify-icon-color-error: var(--red-300);
+  --toastify-color-progress-error: var(--red-300);
```
liza-veis: let's not complicate logic with variables and just use them explicitly, without additional variables for each component
```suggestion
```

frontend/src/libs/packages/store/middlewares/middlewares.ts
```diff
@@ -0,0 +1,21 @@
...
+import { showErrorNotification } from '#slices/notifications/actions.js';
+
+const handleError: Middleware = (api) => {
+  return (next) => {
+    return (action: AnyAction) => {
+      const dispatch: AppDispatch = api.dispatch;
+      if (isRejected(action) && action.error.message) {
+        dispatch(showErrorNotification(action.error.message));
+        NotificationService.error(action.error.message);
+      }
```
liza-veis: 1. next is out dispatch
2. we should call Notification service inside action
3. let's rename notifications slice to app slice, and method to notify (it will accept not only errors
```suggestion
const handleError: Middleware = () => {
  return (next: AppDispatch) => {
    return (action: AnyAction) => {
      if (isRejected(action) && action.error) {
        dispatch(appActions.notify(action.error.message ?? 'Something went wrong'));
      }
```

frontend/src/libs/packages/store/store.package.ts
liza-veis: please use concat here as @HIMDanny said

frontend/src/slices/notifications/actions.ts
```diff
@@ -0,0 +1,5 @@
+import { createAction } from '@reduxjs/toolkit';
+
+const showErrorNotification = createAction<string>('showErrorNotification');
```
liza-veis: Rename it to notify, and also we should get notification service from extra and use it like here
```
 const { notification } = extra;

  notification[type](message);
```

frontend/src/slices/notifications/notification.slice.ts
```diff
@@ -0,0 +1,28 @@
...
+  initialState,
+  reducers: {},
+  extraReducers: (builder) => {
+    builder.addCase(
+      showErrorNotification,
+      (state, action: PayloadAction<string>) => {
+        state.message = action.payload;
+      },
+    );
+  },
```
liza-veis: We don't need to do anything in store with this notification

frontend/src/libs/packages/services/notification-service.ts
```diff
@@ -0,0 +1,9 @@
+import { toast } from 'react-toastify';
+
+const NotificationService = {
```
liza-veis:
```suggestion
const Notification = {
```

frontend/src/libs/packages/services/notification-service.ts
```diff
@@ -0,0 +1,9 @@
...
+  error(message: string): void {
```
liza-veis: Let's create private show method, that will call toast with ` toast(message, parameters);` and other methods like error will pass message and type in parameters, we will need to create  a enum for notification type

liza-veis (changes requested):

OlesiaTr (commented):
Except written above issues - everything looks good to me! Good job!

OlenaYatseniuk (dismissed):
I think you just need to make all request changes. 
You did a great job!

frontend/src/slices/notifications/actions.ts
juanc-escobar: I apologize for any confusion, and I appreciate your efforts to explain how to implement the requested changes. To ensure I'm aligning with the intended modifications, I'd like to request a bit more context regarding the purpose of these changes. This will help me proceed in the right direction.

The comment you provided is from the file located at frontend/src/slices/notifications/actions.ts. Based on this, I assume that the proposed changes should be applied within this file. To utilize the suggested code within action.ts, it seems necessary to incorporate createAsyncThunk in order to gain access to additional functionalities.

Even though notifications may not be asynchronous now and it may be an overkill for the current implementation I don't have the full scope of error handling and what would it do for the final project. regardless of that I did it and it works but the thing is I'm not sure if this is what you wanted to change for that piece of code:

`import { createAsyncThunk } from '@reduxjs/toolkit';
import { type AsyncThunkConfig } from '#libs/types/async-thunk-config.type.js';
import { NotificationType } from '#libs/packages/services/libs/enums/notification-type.enum.js';

export const notify = createAsyncThunk<void, { message: string; type: NotificationType }, AsyncThunkConfig>(
  'notifications/notify',
  ({ message, type }, { extra }) => {
    const { notifications } = extra;
    notifications[type](message);
  }
);`

if this is not what you expect from the changes please let me know being a bit more specific on the intended change so I can have a better understanding and implement what is required from me.

frontend/src/slices/notifications/notification.slice.ts
juanc-escobar: Ok, I have doubts about this one I was thinking that we need to maybe save the state on the store for some future implementations but I don't know the full scope of the error handling system, so should I delete the file?
I'm confused because the other comments say:

let's rename notifications slice to app slice, and method to notify (it will accept not only errors)

so should I eliminate the notification slice (app slice) and not manage states or what should it do?

BrianSammit (dismissed):
With the changes requested should be good.
💪 👍

frontend/src/slices/notifications/actions.ts
BrianSammit: Check my PR hope gives you an idea of how to handle this.
https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/62/files#diff-313760f1b0a3c09ee85709e33b42bae2f95cfab7476f736b107b1f7065d15022

frontend/src/slices/notifications/notification.slice.ts
BrianSammit: The slice is not necessary if we don't need to manage the state.

`If you have an action that doesn't update the store but still needs to be dispatched, you don't necessarily need to create a separate slice for it. Slices are primarily used to manage state and associated logic within the Redux store. If your action doesn't affect the state, it might be better suited as a standalone action creator outside of a slice.`

What I did was follow the structure of the other slices as a reference but leave the slice empty. So it could be use for future state if needed.

frontend/src/slices/notifications/actions.ts
liza-veis: Yes, please check pr above and let me know if you still will have a question

frontend/src/libs/packages/store/store.package.ts
dmytro-kucherenko: In pr #62 on mobile after adding concat the linter error appears, maybe the same will happen here

juanc-escobar: I have tried but I'm not able to the handleError middleware within the concat method in your Redux store setup presents an error message that indicates a type mismatch in the middleware array. The problem seems to arise from the incompatibility of types. since the type includes more middleware than just ThunkMiddleware, which is what is expected, I have tried implementing custom types and searched for solutions but nothing has solved it yet

juanc-escobar: I have tried but I'm not able to, the handleError middleware within the concat method in the Redux store setup presents an error message that indicates a type mismatch in the middleware array. The problem seems to arise from the incompatibility of types. since the type includes more middleware than just ThunkMiddleware, which is what is expected, I have tried implementing custom types and searched for solutions but nothing has solved it yet

iamAlinaaa (dismissed):
great job as for me)) but mind all requests for changes and check are changes done

frontend/src/libs/packages/store/store.package.ts
liza-veis: ok, then let's leave it for now

frontend/src/libs/components/toaster/toaster.tsx
```diff
@@ -0,0 +1,9 @@
+import 'react-toastify/dist/ReactToastify.css';
+
+import { ToastContainer } from 'react-toastify';
+
+const Toaster: React.FC = () => {
```
liza-veis:
```suggestion
const Toast: React.FC = () => {
```

ossavemu (dismissed):
Nice work! The completed error system will help us a lot to those of us who are in fullstack.

frontend/src/libs/packages/services/notification.ts
```diff
@@ -0,0 +1,30 @@
...
+    show(message, NotificationType.INFO);
+  },
+};
+
+function show(
+  message: string,
+  type: (typeof NotificationType)[keyof typeof NotificationType],
+): void {
+  toast(message, { type });
+}
```
liza-veis:
```suggestion
class Notification {
 public [NotificationType.ERROR](message: string): void {
    this.show(message, 'error');
  },

  public [NotificationType.SUCCESS](message: string): void {
    this.show(message, 'success');
  },

   public [NotificationType.WARNING](message: string): void {
    this.show(message, 'warning');
  },

 public [NotificationType.INFO](message: string): void {
    this.show(message, 'info');
  },
  
 private show(
    message: string,
    type: ValueOf<typeof NotificationType>,
  ): void {
    toast(message, { type });
  }
};
```

frontend/src/libs/packages/store/middlewares/middlewares.ts
```diff
@@ -0,0 +1,24 @@
...
+import { appActions } from '#slices/notifications/notification.js';
+
+const handleError: Middleware = ({ dispatch }) => {
+  return (next) => {
+    return (action: AnyAction) => {
+      if (isRejected(action) && action.error.message) {
+        dispatch(
+          appActions.notify({
+            message: action.error.message,
+            type: NotificationType.ERROR,
```
liza-veis:
```suggestion
      if (isRejected(action) && action.error) {
        dispatch(
          appActions.notify({
            message: action.error.message ?? 'Something went wrong',
            type: NotificationType.ERROR,
```

frontend/src/libs/packages/store/store.package.ts
```diff
@@ -56,6 +63,7 @@ class Store {
       authApi,
       userApi,
       storage,
+      notifications: Notification,
```
liza-veis:
```suggestion
      notification,
```

frontend/src/slices/notifications/actions.ts
```diff
@@ -0,0 +1,40 @@
+import { createAction } from '@reduxjs/toolkit';
+
+import { NotificationType } from '#libs/packages/services/libs/enums/enums.js';
+import { Notification } from '#libs/packages/services/notification.js';
+
+import { name as sliceName } from './app.slice.js';
+
+type NotificationPayload = {
```
liza-veis: move it to libs/types/types

frontend/src/slices/notifications/actions.ts
```diff
@@ -0,0 +1,40 @@
...
+  type: (typeof NotificationType)[keyof typeof NotificationType];
```
liza-veis:
```suggestion
  type: ValueOf<typeof NotificationType>;
```

frontend/src/slices/notifications/actions.ts
```diff
@@ -0,0 +1,40 @@
...
+      }
+      case NotificationType.INFO: {
+        Notification.info(message);
+        break;
+      }
+    }
+
+    return { payload: notificationPayload };
+  },
+);
```
liza-veis:
```suggestion
const notify = createAsyncThunk<undefined, NotificationPayload, AsyncThunkConfig>(
  `${sliceName}/notify`,
  (payload, { extra }) => {
    const { notification } = extra;
    const { type, message } = payload;
    
    notification[type](message);
   },
);
```

frontend/src/slices/notifications/app.slice.ts
```diff
@@ -0,0 +1,11 @@
+import { createSlice } from '@reduxjs/toolkit';
```
liza-veis: remove this slice and rename folder from notifications to app

frontend/src/libs/packages/services/notification.ts
```diff
@@ -0,0 +1,30 @@
+import { toast } from 'react-toastify';
```
liza-veis: path should be `libs/packages/notification/notification.package.ts`, not `libs/packages/services/notification.ts`

and then we need to create an instance of this class in `libs/packages/notification/notification.ts` and re-export it as well as NotificationType

frontend/src/libs/packages/services/libs/enums/notification-type.enum.ts
```diff
@@ -0,0 +1,8 @@
+const NotificationType = {
```
liza-veis: should be in `libs/packages/notification/libs/enums/notification-type.enum.ts`

liza-veis (changes requested):

frontend/src/libs/packages/notification/notification.package.ts
```diff
@@ -0,0 +1,29 @@
...
+    this.show(message, 'success');
+  }
+
+  public [NotificationType.WARNING](message: string): void {
+    this.show(message, 'warning');
+  }
+
+  public [NotificationType.INFO](message: string): void {
+    this.show(message, 'info');
+  }
```
v0ldemar01:
```suggestion
  public [NotificationType.ERROR](message: string): void {
    this.show(message, NotificationType.ERROR;
  }

  public [NotificationType.SUCCESS](message: string): void {
    this.show(message, NotificationType.SUCCESS);
  }

  public [NotificationType.WARNING](message: string): void {
    this.show(message, NotificationType.WARNING);
  }

  public [NotificationType.INFO](message: string): void {
    this.show(message, NotificationType.INFO);
  }
```

frontend/src/libs/packages/notification/libs/types.ts/notification-payload.type.ts
v0ldemar01, line 1: Remove .ts from folder name

v0ldemar01 (changes requested):
Nice job, last comments to be fixed

v0ldemar01 (dismissed):

frontend/src/libs/packages/notification/notification.package.ts
```diff
@@ -0,0 +1,29 @@
+import { toast } from 'react-toastify';
+
+import { type ValueOf } from '#libs/types/types.js';
+
+import { NotificationType } from './libs/enums/notification-type.enum.js';
```
fshabanov:
```suggestion
import { NotificationType } from './libs/enums/enums.js';
```

frontend/src/slices/app/actions.ts
```diff
@@ -0,0 +1,17 @@
+import { createAsyncThunk } from '@reduxjs/toolkit';
+
+import { type NotificationPayload } from '#libs/packages/notification/libs/types/types.js';
+import { type Notification } from '#libs/packages/notification/notification.package.js';
+import { type AsyncThunkConfig } from '#libs/types/types.js';
+
+const notify = createAsyncThunk<unknown, NotificationPayload, AsyncThunkConfig>(
+  'app/notify',
+  (payload, { extra }) => {
+    const notification: Notification = extra.notification;
```
fshabanov:
```suggestion
    const { notification } = extra;
```

frontend/src/slices/app/actions.ts
```diff
@@ -0,0 +1,17 @@
+import { createAsyncThunk } from '@reduxjs/toolkit';
+
+import { type NotificationPayload } from '#libs/packages/notification/libs/types/types.js';
```
fshabanov:
```suggestion
import { type NotificationPayload } from '#libs/packages/notification/notification.js';
```
Please, check other imports as well

fshabanov (changes requested):
Nice job! Please, fix these last comments

frontend/src/slices/app/app-notification.ts
```diff
@@ -0,0 +1,7 @@
+import { notify } from './actions.js';
+
+const allActions = {
+  notify,
+};
+
+export { allActions as appActions };
```
v0ldemar01: Maybe leave appActions only

frontend/src/assets/css/variables.scss
```diff
@@ -1,18 +1,8 @@
...
-  --violet-100: #e5e1ff;
-  --pink-200: #e1a7d5;
-  --red-300: #d51a52;
   --transparent: #00000000;
   --white-alpha-20: #ffffff33;
   --white-alpha-80: #ffffff80;
   --blue-alpha-30: #527deb4d;
+  --red-error: #d51a52;
+  --black-text: #302e36;
 }
```
liza-veis: return changes in this file back and use variables from it
```suggestion
:root {
  --white: #ffffff;
  --shadow: 0 2px 4px #0000001a;
  --black: #000000;
  --gray-100: #e2e3e5;
  --gray-200: #6c707a;
  --gray-300: #302e36;
  --blue-100: #ebf3ff;
  --blue-200: #acb4d0;
  --blue-300: #507ceb;
  --violet-100: #e5e1ff;
  --pink-200: #e1a7d5;
  --red-300: #d51a52;
  --transparent: #00000000;
  --white-alpha-20: #ffffff33;
  --white-alpha-80: #ffffff80;
  --blue-alpha-30: #527deb4d;
}
```

frontend/src/libs/packages/store/middlewares/middlewares.ts
```diff
@@ -0,0 +1,27 @@
...
+import { NotificationType } from '#libs/packages/notification/libs/enums/enums.js';
+import { type AppDispatch } from '#libs/packages/store/libs/types/types.js';
+import { appActions } from '#slices/app/app-notification.js';
+
+const handleError: Middleware<unknown, unknown, AppDispatch> = ({
+  dispatch,
+}) => {
+  return (next) => {
+    return (action: AnyAction) => {
+      if (isRejected(action)) {
```
liza-veis, line 13:
```suggestion
      if (isRejected(action) && action.error) {
```

liza-veis (changes requested):

frontend/src/libs/packages/store/middlewares/middlewares.ts
juanc-escobar: I previously had it set up in the same manner, but without the conditional part (**?? 'Something went wrong'**). The issue is that the ESLint rules don't permit both configurations:

At line 14, column 22, there's an error flagged: "Unnecessary conditional," with the message advising the left-hand side of the ?? operator to potentially be null or undefined, as per the @typescript-eslint/no-unnecessary-condition rule.
Because we introduced the conditional check '**&& action.error**,' we will never utilize the ?? 'Something went wrong' part. TypeScript interprets it as an unnecessary condition.

Could you please let me know which solution you would prefer me to implement?

liza-veis: let's use it without `?? 'Something went wrong'` then

juanc-escobar: I previously had it set up in the same manner, but without the conditional part (?? 'Something went wrong'). The issue is that the ESLint rules don't permit both configurations:

  13:33  error  Unnecessary conditional, value is always truthy  @typescript-eslint/no-unnecessary-condition
Because we introduced the conditional check '&& action.error,' we will never utilize the ?? 'Something went wrong' part. TypeScript interprets it as an unnecessary condition.

Could you please let me know which solution you would prefer me to implement?

frontend/src/libs/packages/store/store.package.ts
```diff
@@ -23,6 +26,7 @@ type ExtraArguments = {
   authApi: typeof authApi;
   userApi: typeof userApi;
   storage: typeof storage;
+  notification: Notification;
```
liza-veis: We don't need to export Notification
```suggestion
  notification: typeof notification;
```

liza-veis (changes requested):

frontend/src/libs/packages/store/middlewares/middlewares.ts
juanc-escobar: i will update this, i saw your comment after committing.

juanc-escobar: Hi, I'm back again! After all the tweaks we've made, if I take out "action.error," I get this error message:

Type 'string | undefined' is not assignable to type 'string'.
 Type 'undefined' is not assignable to type 'string'.ts(2322)
notification-payload.type.ts(7, 3): The expected type comes from property 'message' which is declared here on type 'NotificationPayload'

We're expecting a string, but the message might sometimes be undefined. To fix this, one way is to change "&& action.error" to "&& action.error.message." or do a triple-check. If the message is there, it'll pop up. However, if there's a glitch without a message, we won't catch it.

Now, about "&& action.error": if we keep it and the error message doesn't exist, we'll end up with an empty space. In this case, I'm thinking it might be better to stick with the "?? 'Something went wrong'" condition and drop the "&& action.error." If something goes wrong and the action gets rejected without triggering an "action.error," at least we'll display "Something went wrong" so users know there's an issue. Plus, if the error does get added to the rejection later on, we'll be all set to show the error message.

Of course, that's just my opinion. I'm sure there are other ways to go about it, but we're kinda of limited by the linter and type restrictions. For now, I'll go ahead and push the code like this. Take your time to read through this and let me know which solution you're leaning toward, or if you have a better solution let me know and I will implement it as soon as possible.

liza-veis (dismissed):

v0ldemar01 (dismissed):

fshabanov (dismissed):

liza-veis (approved):

Issue #12, opened 2023-08-19
Sign In


liza-veis:
- [x] Model (get user for sign-in)
- [x] Service (sign-in user flow)
- [x] Controller (sign-in route)
- [x] Validation (shared)

OlesiaTr:
Hello! Is there a document where I can look up responses or could you, please, tell me what format does signIn response have?

liza-veis:
The same response as for sign up, just be aware that name property will be added in Sign Up #14 and token in Authorization token #16, but this doesn't impact you
```
{
  "user": {
      "id": number,
      "email": string,
      "createdAt": Date,
      "updatedAt": Date,
   }
}
```

OlesiaTr:
I don't add fullName?

liza-veis:
Yes, it will be added in sign-up task

Xxandder:
In sign ip there is no need to return fullName field? @Liza-Veis

liza-veis:
@Xxandder Yes, will also need to return it, we will use returned user as authenticated user

Xxandder:
Sorry, I wrote sign 'ip', so you said about sign in?)

liza-veis:
yes, we will have the same interface for user in sign-up/sign-in/authenticated-user

PR #37, opened 2023-08-21
cp-16: add jwt

vladyslavdobychin:
- Implemented basic JWT authorisation for Sign Up route.
- Was not able to do the same for login route since it was not implemented yet, waiting for [issue 12](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/issues/12)
- Submitting draft pull request


shared/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "../tsconfig.json",
-  "include": ["src"],
+  "include": ["src", "build"],
```
v0ldemar01: Are you sure that it is needed?

backend/src/packages/users/user.service.ts
```diff
@@ -1,3 +1,6 @@
+import { SignJWT } from 'jose';
```
fshabanov: Let's create a separate jwt service, since we will use it for auth plugin as well.

backend/src/packages/users/users.ts
```diff
@@ -6,7 +7,8 @@ import { UserRepository } from './user.repository.js';
 import { UserService } from './user.service.js';
 
 const userRepository = new UserRepository(UserModel);
-const userService = new UserService(userRepository);
+const config = new BaseConfig(logger);
```
fshabanov: Import config from #libs/packages/config/config.ts, do not create a new instance

frontend/package-lock.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "frontend",
```
fshabanov: Why do you add 'jose' to the frontend?

frontend/src/packages/auth/auth-api.ts
```diff
@@ -32,8 +32,12 @@ class AuthApi extends BaseHttpApi {
         hasAuth: false,
       },
     );
+    const parsedResponse = await response.json<UserSignUpResponseDto>();
 
-    return await response.json<UserSignUpResponseDto>();
+    // Saving the token into the local storage
```
fshabanov: There's no need for this comment

fshabanov (changes requested):

shared/tsconfig.json
vladyslavdobychin: Initially, I added this change to modify UserSignUpResponseDto to add the token,  but I would get an error: 
`Parsing error: Cannot read file '/users/vlad_d/desktop/bsa-2023-calmpal/shared/shared/tsconfig.json'.`
The only solution I found is to include the build directory that contained the UserSignUpResponseDto type. 
If there is abetter way to fix it I'll gladly change it.

backend/src/packages/users/user.service.ts
vladyslavdobychin: Okay, will do, thank you!

frontend/package-lock.json
vladyslavdobychin: Initially, I thought that I'll need it, which was a mistake. I thought I removed it in my latest [commit](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/commit/bca895e22208e00baa9bfd86aeb3a1d2db8ec37e)

shared/tsconfig.json
v0ldemar01: That is the issue with eslint plugin and it won't influenced to linting the project. Typescript should check only src/ folder)

vladyslavdobychin: So I can safely remove, then? Okay, thank you!

backend/package-lock.json
v0ldemar01, line 1: Backend should not include package-lock.json? Are you using npm i -w backend in root?

v0ldemar01 (changes requested):

backend/package-lock.json
vladyslavdobychin: Shouldn't it? Well, I did a regular setup described in the Readme: 
1. npm install
2. npx simple-git-hooks
3. npm run migrate:dev -w backend
4. npm run start:dev -w backend
5. npm run start:dev -w frontend

v0ldemar01: This flow doesn't include the installation of new package to project)

vladyslavdobychin: Let me try reiterate for myself, because I think I lost you. 
So, the backend directory should not include the package-lock.json, but for some reason, mine - does?

v0ldemar01: We are having the npm workspaces approach, it provides the solution to put a tree of dependencies in one place to reduce complexity with installing and managing ones

vladyslavdobychin: Okay, I think I'm getting there. When I was installing jose, I went to back end directory and just did npm install jose there. Do I get it right that this was not a desirable way and I had to do this somehow differently, using workspaces from the root directory?

v0ldemar01: You are right)

vladyslavdobychin: At this point, do I need to re-install jose but via workspaces?

liza-veis: yes

backend/src/packages/auth/jwt.service.ts
```diff
@@ -0,0 +1,17 @@
...
+class JWTService {
+  private secret: Buffer;
+
+  public constructor(secret: string) {
+    this.secret = Buffer.from(secret, 'utf8');
+  }
+
+  public async signJWT(payload: Record<string, number>): Promise<string> {
+    return await new SignJWT(payload)
+      .setProtectedHeader({ alg: 'HS256' })
```
v0ldemar01: Let's use alg as secret one

v0ldemar01 (changes requested):

backend/src/packages/auth/jwt.service.ts
v0ldemar01: Provide as ALG: 'HS256' in base config but with default value

vladyslavdobychin: You mean switching for HS to RS?

vladyslavdobychin: > Provide as ALG: 'HS256' in base config but with default value

Will that be good enough? 

Screenshot 2023-08-22 at 12 26 29

v0ldemar01: No, as a separate declaration, so AUTH contains JWT_SECRET and AIG

vladyslavdobychin: > No, as a separate declaration, so AUTH contains JWT_SECRET and AIG

You think that will be better? 
Screenshot 2023-08-22 at 12 32 40
Screenshot 2023-08-22 at 12 32 53

vladyslavdobychin: Will rename JWT_ALGORITHM into JWT_ALG ofcourse

vladyslavdobychin: > No, as a separate declaration, so AUTH contains JWT_SECRET and AIG

I hope that will do: 

Screenshot 2023-08-22 at 12 55 21

Screenshot 2023-08-22 at 12 55 46
Screenshot 2023-08-22 at 12 55 56

v0ldemar01: It would be
`
JWT: {
    SECRET_KEY: {//},
    ALG: {//}
}
`

v0ldemar01: And ALG with default value so .env and .env.example won't contain ALG record

vladyslavdobychin: > And ALG with default value so .env and .env.example won't contain ALG record

Okay, got it, hope it'll work: 

Screenshot 2023-08-22 at 13 08 58

backend/src/libs/packages/config/libs/types/environment-schema.type.ts
```diff
@@ -13,6 +13,12 @@ type EnvironmentSchema = {
     POOL_MIN: number;
     POOL_MAX: number;
   };
+  JWT: {
+    SECRET_KEY: string;
+    ALG: {
+      JWT_ALG: string;
+    };
```
HIMDanny: Shouldn't it be like so?

```suggestion
    ALG: string;
```

vladyslavdobychin: I think you are right, I attached the latest implementation [here](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/37#discussion_r1301414975)

backend/src/packages/auth/jwt.service.ts
v0ldemar01: The last thing is to replace `alg` prop in `ALG` object to `default`

backend/.env.example
```diff
@@ -12,3 +12,10 @@ DB_CONNECTION_STRING=[db_client]://[db_username]:[db_user_password]@localhost:[d
 DB_DIALECT=pg
 DB_POOL_MIN=2
 DB_POOL_MAX=10
+
+#
+# AUTH
+#
+SECRET_KEY=secret
```
HIMDanny: Maybe we should be more precise here and name it JWT_SECRET_KEY, like in other variables?

backend/src/packages/auth/jwt.service.ts
vladyslavdobychin: Understood, just to confirm: 

Screenshot 2023-08-22 at 13 15 42
Screenshot 2023-08-22 at 13 15 52

backend/.env.example
vladyslavdobychin: I think we will need to ask @v0ldemar01 about that

v0ldemar01: https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/39#discussion_r1301130378

backend/src/packages/auth/jwt.service.ts
v0ldemar01: Also don't forget to import ALG to JWTService)

HIMDanny: can we just import it directly, or it is better to pass alg through constructor like a secret property?

vladyslavdobychin:
@HIMDanny was very kind to provide me his implementation of decoder for JWTService in [2a0209b](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/37/commits/2a0209b15a19b311a185e17552997d239d85cd87)

frontend/src/packages/auth/auth-api.ts
```diff
@@ -32,8 +32,11 @@ class AuthApi extends BaseHttpApi {
         hasAuth: false,
       },
     );
+    const parsedResponse = await response.json<UserSignUpResponseDto>();
 
-    return await response.json<UserSignUpResponseDto>();
+    localStorage.setItem('token', parsedResponse.token);
```
v0ldemar01: We already have a storage service as abstraction for localStorage, please reuse it

v0ldemar01 (changes requested):

frontend/src/packages/auth/auth-api.ts
v0ldemar01: And 'token' -> StorageKey.TOKEN

vladyslavdobychin: > And 'token' -> StorageKey.TOKEN

That's also very important note, I wouldn't have thought about it, thank you!

vladyslavdobychin:
- Modified the SignUp method in the latest [commit](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/37/commits/78e149b8a68cead26a64819abe75b97c5f68d3ba) to use the Storage Service instead of plain localStorage. 
- Ran into an unexpected issue with jwtService, looks like a case of circular dependencies. Looking for the best way to resolve it. 
Screenshot 2023-08-22 at 17 51 56

OlenaYatseniuk (dismissed):
It looks good to me. Just need to change the code to fit the coaches' suggestions.

backend/package-lock.json
```diff
@@ -0,0 +1,1469 @@
+{
```
fshabanov: We shouldn't have package-lock.json in backend folder

backend/src/packages/auth/jwt.service.ts
```diff
@@ -0,0 +1,30 @@
...
+  public constructor({ secret, alg }: Constructor) {
+    this.secret = Buffer.from(secret, 'utf8');
+    this.alg = alg;
+  }
+
+  public async signJWT(payload: Record<string, number>): Promise<string> {
+    return await new SignJWT(payload)
+      .setProtectedHeader({ alg: this.alg })
+      .sign(this.secret);
+  }
```
fshabanov:
```suggestion
  public signJWT(payload: Record<string, number>): Promise<string> {
    return new SignJWT(payload)
      .setProtectedHeader({ alg: this.alg })
      .sign(this.secret);
  }
```

frontend/package-lock.json
```diff
@@ -0,0 +1,13 @@
...
+  "version": "0.0.1",
```
fshabanov: We shouldn't have package-lock.json in frontend folder

frontend/src/packages/auth/auth-api.ts
```diff
@@ -16,8 +17,11 @@ type Constructor = {
 };
 
 class AuthApi extends BaseHttpApi {
+  private authStorage: Storage;
```
fshabanov:
```suggestion
  private storage: Storage;
```

shared/src/packages/users/libs/types/user-sign-up-response-dto.type.ts
```diff
@@ -1,6 +1,7 @@
 type UserSignUpResponseDto = {
   id: number;
   email: string;
+  token: string;
```
fshabanov, line 5: Let's send token separately. Something like this:
{
    user: UserSignUpResponseDto;
    token: string
}

Some other guys already implemented it like this, please, check their PRs

fshabanov (changes requested):

backend/package-lock.json
vladyslavdobychin: Got it, will work on it

shared/src/packages/users/libs/types/user-sign-up-response-dto.type.ts
vladyslavdobychin: Got it, will try to find it

v0ldemar01:
> * Modified the SignUp method in the latest [commit](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/37/commits/78e149b8a68cead26a64819abe75b97c5f68d3ba) to use the Storage Service instead of plain localStorage.
> * Ran into an unexpected issue with jwtService, looks like a case of circular dependencies. Looking for the best way to resolve it.
> 
> Screenshot 2023-08-22 at 17 51 56

That caused due to import user to auth(which contain jwtService) and import jwt to user. Let's move jwt as separate package to libs/packages/

vladyslavdobychin:
> That caused due to import user to auth(which contain jwtService) and import jwt to user. Let's move jwt as separate package to libs/packages/

@v0ldemar01 I just tried moving JWT instance to libs/packages in the latest [commit](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/37/commits/7a780c70b99fab463c0a3842307cf189d3137ffc). Do you mind taking a look?

v0ldemar01:
> > That caused due to import user to auth(which contain jwtService) and import jwt to user. Let's move jwt as separate package to libs/packages/
> 
> @v0ldemar01 I just tried moving JWT instance to libs/packages in the latest [commit](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/37/commits/7a780c70b99fab463c0a3842307cf189d3137ffc). Do you mind taking a look?

Partially yes, but also move the service too, so jwt package will include jwt service and expose from package

vladyslavdobychin:
> > > That caused due to import user to auth(which contain jwtService) and import jwt to user. Let's move jwt as separate package to libs/packages/
> > 
> > 
> > @v0ldemar01 I just tried moving JWT instance to libs/packages in the latest [commit](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/37/commits/7a780c70b99fab463c0a3842307cf189d3137ffc). Do you mind taking a look?
> 
> Partially yes, but also move the service too, so jwt package will include jwt service and expose from package

I hope [this](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/37/commits/766a0ddda1c680732e128401225db355216533bf) will be better

backend/src/packages/users/user.service.ts
```diff
@@ -38,7 +41,13 @@ class UserService implements Service {
       }),
     );
 
-    return item.toObject();
+    const userObject = item.toObject();
```
v0ldemar01: Please follow B17 rule

backend/src/packages/users/user.service.ts
```diff
@@ -10,9 +11,11 @@ import {
 
 class UserService implements Service {
   private userRepository: UserRepository;
+  private jwtService: JWTService;
 
-  public constructor(userRepository: UserRepository) {
+  public constructor(userRepository: UserRepository, jwtService: JWTService) {
```
v0ldemar01: As we are using the approach of one param to constructor, let's proceed with it here

v0ldemar01 (changes requested):

backend/src/packages/users/user.service.ts
vladyslavdobychin: I think I did it, that's a great Idea, thank you!

v0ldemar01 (dismissed):
`const jwt = await this.jwtService.signJWT({ userId: user.id });` Small improvement jwt -> token, and please fix Farid's [comment](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/37/files#r1301858018)

frontend/src/packages/auth/auth-api.ts
vladyslavdobychin: I tried making this change and replace authStorage with storage, but it would cause an error: 
`Class 'AuthApi' incorrectly extends base class 'BaseHttpApi'. Types have separate declarations of a private property 'storage'.`
As far as I could understand, it happens because of BaseHttpApi already having its own storage property, the best solution I could come up with is to change the property **storage** in BaseHttpApi from private to protected. And completely remove assignment of private storage in auth-api. But honestly, I am not sure if it is a good idea to change the base class like this: 

Screenshot 2023-08-23 at 12 19 19

Screenshot 2023-08-23 at 12 19 34

Screenshot 2023-08-23 at 12 19 45

v0ldemar01: let's stay with protected directive

vladyslavdobychin: Does it mean that I should keep authStorage the way it was?

v0ldemar01: Stay with protected storage one) let's have only one storage

vladyslavdobychin: Oh, yes, my apologies

v0ldemar01 (dismissed):

backend/package-lock.json
vladyslavdobychin: Hi, @v0ldemar01  
since you are already here, could you please have a look at what Farid requested before? I think I fixed it

shared/src/packages/users/libs/types/user-sign-up-response-dto.type.ts
```diff
@@ -1,6 +1,11 @@
-type UserSignUpResponseDto = {
+type UserAuthResponseDto = {
```
liza-veis: Move it to a separate file, we will use it for sign-in and get authenticated-user too

frontend/src/packages/auth/auth-api.ts
```diff
@@ -32,8 +33,11 @@ class AuthApi extends BaseHttpApi {
         hasAuth: false,
       },
     );
+    const parsedResponse = await response.json<UserSignUpResponseDto>();
 
-    return await response.json<UserSignUpResponseDto>();
+    await this.storage.set(StorageKey.TOKEN, parsedResponse.token);
```
liza-veis: this logic should be in sign up action

liza-veis (changes requested):

shared/src/packages/users/libs/types/user-sign-up-response-dto.type.ts
vladyslavdobychin: Will do!

v0ldemar01 (dismissed):

fshabanov (dismissed):

frontend/src/packages/auth/auth-api.ts
vladyslavdobychin: My apologies, it took me longer than I expected. I couldn't figure out how to do this. The only solution I could come up with is to pass the store as an extra argument to signUp thunk, so I could use it from there for saving the token. 
Here is the latest [commit](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/37/commits/a9fba047eb3b05db6e5e84c33ac409a800a6a395)

v0ldemar01 (approved):

liza-veis (approved):

vladyslavdobychin:
@v0ldemar01 @fshabanov @Liza-Veis 
I am very grateful for all of your patience and support throughout this task. I do mean it and I do appreciate it.  
You are great coaches.

Issue #29, opened 2023-08-19
Authenticated user flow


liza-veis:
- [x] Service (get authenticated user)
- [x] Controller (authenticated-user endpoint)

VINAAAYYY:
Hi @Liza-Veis I want to contribute to this issue

liza-veis:
Hey @VINAAAYYY!  Thank you for your interest in the project, but, unfortunately, we do not accept contributors yet

OlenaYatseniuk:
Hi, @Liza-Veis!
Could you, please, clarify this?
As far as I understood, in service in getAuthenticatedUser method, I need to use jwtService from pr #37.  I think I'm not responsible to create this in my issue, but I'm stuck with it now. How could I deal with it now? Or maybe am I not right about this method implementation?
Also, I don't know how to test when I finish it, because for now, if I correctly understand I don't have any possibility to have a jwt token when I sign up/sign in.
Sorry, for such confusing questions, but I feel a lack of experience in understanding such a big structure.  
Thank you for your patience!

vladyslavdobychin:
If I followed you correctly, and you need the jwtService, you can already find it implemented in pr https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/37. 
I'm not sure If I can state it on my behalf, but I submitted the jwtService and seemed to me that neither Farid, nor Vladimir had anything to add about it. 
Also Danylo is already using this implementation in his pr. 
But again, depends on what Liza says.

OlenaYatseniuk:
Thank you for your answer! I will review your pr again to get it.

вт, 22 авг. 2023 г. в 17:04, Vladyslav Dobychin ***@***.***>:

liza-veis:
Hey) let's wait with this task until pr https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/37 is merged, I will assign a new task to you

OlenaYatseniuk:
Got it, thanks!

HIMDanny:
Hey, @Liza-Veis! We already have the user in the request with the help of the authorization plugin. I can just use the user service in the auth controller to find the user by id, can I? Or do you want me to add the method in the auth service that will call the find by id method from the user service and work only with auth service in the auth controller? I hope I wrote it clearly)

liza-veis:
Hey) 
> Or do you want me to add the method in the auth service that will call the find by id method from the user service and work only with auth service in the auth controller?

[...]

Issue #18, opened 2023-08-19
Protected routing


liza-veis:
- [x] Add Protected Route wrapper component
- [x] Redirect not signed in users from protected routes to Sign In page

Issue #14, opened 2023-08-19
Sign Up


liza-veis:
- [x] Migration (create user_details table, should contain all fields from the db schema except avatar_id and survey)
- [x] Model (user_details saving)
- [x] Service (user & user_details data saving flow)
- [x] Controller (sign-up route)
- [x] Validation (shared)

Xxandder:
Hello, Could you, please, tell me what format does signUp response have? Does it need to have fields from user_details?

liza-veis:
Hey) the response should be the following

```
{
  "user": {
    "id"
    "createdAt"
    "updatedAt"
    "email"
    "fullName"
  }
}
```
So, you need to join userDetails to user and then remap user on repository level so that it will return object without nested userDetails

Xxandder:
Hello, could you, please, tell me what I am supposed to do in signUp route? It already works and cathches errors

Xxandder:
And one more question about the response format. Yevhenii need token in reaponse so maybe I need to send response like this?
{
"message":
    "user": {
      "id"
      "createdAt"
      "updatedAt"
      "email"
      "fullName"
    }
    "token"
}
I added "message" corresponding to a comments in the project:


![Image](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/111736110/93850f26-db84-4631-870b-688aec07fb39)

liza-veis:
Token will be added in this task Authorization Token #16

Xxandder:
Thanks, got it about the token. and about authController what I am supposed to do in signUp route? It already works and cathches errors

liza-veis:
if it works, then nothing)

Xxandder:
Ok, thank you for the aswer

Xxandder:
What error status do we need in register error when user is already exists ?

liza-veis:
@Xxandder we need to return 400 - Bad Request

oksana-mlynska:
@Xxandder Now the response message after signing up looks like this:
{
    "user": {
        "id": 
        "email": 
        "fullName": 
        "createdAt": 
        "updatedAt": 
        "isSurveyCompleted": 
        "passwordHash": 
        "passwordSalt": 
    },
    "token":
}

it's because more changes and new tasks were applied after finishing this issue, right?

oksana-mlynska:
[...]

PR #40, opened 2023-08-21
cp-19: Authenticated user flow

Nanaier:
Can't test the code since we don't have a backend endpoint to do so, so I leave this PR as a draft for now.
This PR depends on this issue: #29

### DIFF
--- a/frontend/src/libs/components/app/app.tsx
+++ b/frontend/src/libs/components/app/app.tsx
@@ -12,9 +12,7 @@ import {
   useAppSelector,
   useEffect,
   useLocation,
-  useNavigate,
 } from '#libs/hooks/hooks.js';
-import { storage, StorageKey } from '#libs/packages/storage/storage.js';
 import { actions as authActions } from '#slices/auth/auth.js';
 import { actions as userActions } from '#slices/users/users.js';
 
@@ -24,11 +22,10 @@ import styles from './styles.module.scss';
 const App: React.FC = () => {
   const { pathname } = useLocation();
   const dispatch = useAppDispatch();
-  const navigate = useNavigate();
   const { authenticatedUserDataStatus } = useAppSelector(({ users, auth }) => ({
     users: users.users,
     dataStatus: users.dataStatus,
-    user: auth.user,
+    authenticatedUser: auth.authenticatedUser,
     authenticatedUserDataStatus: auth.authenticatedUserDataStatus,
   }));
 
@@ -41,16 +38,12 @@ const App: React.FC = () => {
   }, [isRoot, dispatch]);
 
   useEffect(() => {
-    const checkTokenAndFetchUser = async (): Promise<void> => {
-      if (!(await storage.has(StorageKey.TOKEN))) {
-        navigate(AppRoute.SIGN_IN);
-      }
+    void dispatch(authActions.getAuthenticatedUser());
+  }, [dispatch]);
 
-      void dispatch(authActions.getAuthenticatedUser());
-    };
-
-    void checkTokenAndFetchUser();
-  }, [dispatch, navigate]);
+  if (authenticatedUserDataStatus !== DataStatus.FULFILLED) {
+    return <Loader />;
+  }
 
   return (
     <div className={styles['app-container']}>
--- a/frontend/src/libs/components/app/app.tsx
+++ b/frontend/src/libs/components/app/app.tsx
@@ -22,10 +22,7 @@ import styles from './styles.module.scss';
 const App: React.FC = () => {
   const { pathname } = useLocation();
   const dispatch = useAppDispatch();
-  const { authenticatedUserDataStatus } = useAppSelector(({ users, auth }) => ({
-    users: users.users,
-    dataStatus: users.dataStatus,
-    authenticatedUser: auth.authenticatedUser,
+  const { authenticatedUserDataStatus } = useAppSelector(({ auth }) => ({
     authenticatedUserDataStatus: auth.authenticatedUserDataStatus,
   }));
 
--- a/frontend/src/libs/components/app/app.tsx
+++ b/frontend/src/libs/components/app/app.tsx
@@ -46,7 +46,7 @@ const App: React.FC = () => {
       .catch(() => {
         navigate(AppRoute.SIGN_IN);
       });
-  }, []);
+  }, [dispatch, navigate]);
 
   return (
     <div className={styles['app-container']}>
--- a/frontend/src/libs/components/app/app.tsx
+++ b/frontend/src/libs/components/app/app.tsx
@@ -1,7 +1,12 @@
 import home from '#assets/img/home.svg';
 import reactLogo from '#assets/img/react.svg';
-import { Header, Link, RouterOutlet } from '#libs/components/components.js';
-import { AppRoute } from '#libs/enums/enums.js';
+import {
+  Header,
+  Link,
+  Loader,
+  RouterOutlet,
+} from '#libs/components/components.js';
+import { AppRoute, DataStatus } from '#libs/enums/enums.js';
 import {
   useAppDispatch,
   useAppSelector,
@@ -9,7 +14,7 @@ import {
   useLocation,
   useNavigate,
 } from '#libs/hooks/hooks.js';
-import { StorageKey } from '#libs/packages/storage/storage.js';
+import { storage, StorageKey } from '#libs/packages/storage/storage.js';
 import { actions as authActions } from '#slices/auth/auth.js';
 import { actions as userActions } from '#slices/users/users.js';
 
@@ -20,14 +25,12 @@ const App: React.FC = () => {
   const { pathname } = useLocation();
   const dispatch = useAppDispatch();
   const navigate = useNavigate();
-  const { users, dataStatus, user, userDataStatus } = useAppSelector(
-    ({ users, auth }) => ({
-      users: users.users,
-      dataStatus: users.dataStatus,
-      user: auth.user,
-      userDataStatus: auth.dataStatus,
-    }),
-  );
+  const { authenticatedUserDataStatus } = useAppSelector(({ users, auth }) => ({
+    users: users.users,
+    dataStatus: users.dataStatus,
+    user: auth.user,
+    authenticatedUserDataStatus: auth.authenticatedUserDataStatus,
+  }));
 
   const isRoot = pathname === AppRoute.ROOT;
 
@@ -38,14 +41,15 @@ const App: React.FC = () => {
   }, [isRoot, dispatch]);
 
   useEffect(() => {
-    if (!localStorage.getItem(StorageKey.TOKEN)) {
-      navigate(AppRoute.SIGN_IN);
-    }
-    void dispatch(authActions.getUser())
-      .unwrap()
-      .catch(() => {
+    const checkTokenAndFetchUser = async (): Promise<void> => {
+      if (!(await storage.has(StorageKey.TOKEN))) {
         navigate(AppRoute.SIGN_IN);
-      });
+      }
+
+      void dispatch(authActions.getAuthenticatedUser());
+    };
+
+    void checkTokenAndFetchUser();
   }, [dispatch, navigate]);
 
   return (
--- a/frontend/src/libs/components/app/app.tsx
+++ b/frontend/src/libs/components/app/app.tsx
@@ -41,7 +41,7 @@ const App: React.FC = () => {
     void dispatch(authActions.getAuthenticatedUser());
   }, [dispatch]);
 
-  if (authenticatedUserDataStatus !== DataStatus.FULFILLED) {
+  if (authenticatedUserDataStatus === DataStatus.PENDING) {
     return <Loader />;
   }
 
--- a/frontend/src/libs/components/app/app.tsx
+++ b/frontend/src/libs/components/app/app.tsx
@@ -2,7 +2,15 @@ import home from '#assets/img/home.svg';
 import reactLogo from '#assets/img/react.svg';
 import { Header, Link, RouterOutlet } from '#libs/components/components.js';
 import { AppRoute } from '#libs/enums/enums.js';
-import { useAppDispatch, useEffect, useLocation } from '#libs/hooks/hooks.js';
+import {
+  useAppDispatch,
+  useAppSelector,
+  useEffect,
+  useLocation,
+  useNavigate,
+} from '#libs/hooks/hooks.js';
+import { StorageKey } from '#libs/packages/storage/storage.js';
+import { actions as authActions } from '#slices/auth/auth.js';
 import { actions as userActions } from '#slices/users/users.js';
 
 import { Sidebar } from '../sidebar/sidebar.js';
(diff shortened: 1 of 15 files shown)


liza-veis:
Please mention the issue your pr depends on in the description, it will be automatically attached to that issue and we will be able to see all its dependencies

frontend/src/libs/components/app/app.tsx
```diff
@@ -16,6 +33,17 @@ const App: React.FC = () => {
     }
   }, [isRoot, dispatch]);
 
+  useEffect(() => {
+    if (!localStorage.getItem(StorageKey.TOKEN)) {
```
v0ldemar01: Let's use a wrapper service for using localStorage

frontend/src/libs/components/app/app.tsx
```diff
@@ -16,6 +33,17 @@ const App: React.FC = () => {
...
+      navigate(AppRoute.SIGN_IN);
+    }
+    void dispatch(authActions.getUser())
+      .unwrap()
+      .catch(() => {
```
v0ldemar01: Actually, the redirection will navigated on private route abstraction

v0ldemar01 (changes requested):

frontend/src/libs/components/app/app.tsx
liza-veis: we already have it - storage package

frontend/src/libs/components/app/app.tsx
Liza-Veis: agree, for now we just need to get user

frontend/src/packages/auth/auth-api.ts
```diff
@@ -35,6 +36,19 @@ class AuthApi extends BaseHttpApi {
 
     return await response.json<UserSignUpResponseDto>();
   }
+
+  public async getUser(): Promise<UserGetAllItemResponseDto> {
```
liza-veis: Let's create a new type in shared, for now it will have the same interface, but we will change it in Sign Up #14 

```suggestion
  public async getAuthenticatedUser(): Promise<UserAuthResponseDto> {
```

frontend/src/packages/auth/auth-api.ts
```diff
@@ -35,6 +36,19 @@ class AuthApi extends BaseHttpApi {
...
+    const response = await this.load(
+      this.getFullEndpoint(AuthApiPath.USER, {}),
+      {
+        method: 'GET',
+        contentType: ContentType.JSON,
```
liza-veis: We don't send body
```suggestion
```

frontend/src/slices/auth/actions.ts
```diff
@@ -18,4 +18,14 @@ const signUp = createAsyncThunk<
   return authApi.signUp(registerPayload);
 });
 
-export { signUp };
+const getUser = createAsyncThunk<
```
liza-veis: And also rename it everywhere 
```suggestion
const getAuthenticatedUser = createAsyncThunk<
```

frontend/src/slices/auth/auth.slice.ts
```diff
@@ -27,6 +30,16 @@ const { reducer, actions, name } = createSlice({
     builder.addCase(signUp.rejected, (state) => {
       state.dataStatus = DataStatus.REJECTED;
     });
+    builder.addCase(getUser.pending, (state) => {
```
liza-veis: Let's add a space before, to separate this block

shared/src/packages/auth/libs/enums/auth-api-path.enum.ts
```diff
@@ -1,6 +1,7 @@
 const AuthApiPath = {
   ROOT: '/',
   SIGN_UP: '/sign-up',
+  USER: '/user',
```
liza-veis: Endpoint also will be '/authenticated-user'

liza-veis (changes requested):

frontend/src/libs/components/app/app.tsx
```diff
@@ -37,6 +65,19 @@ const App: React.FC = () => {
       <div>
         <RouterOutlet />
       </div>
+      {isRoot && (
```
liza-veis: Add a `// TODO: message comment to remove this`

liza-veis:
Also we already merged ProtectedRoute component that depends on your pr #18. Could you please update it with user? we left a todo comment there

frontend/src/slices/auth/auth.slice.ts
```diff
@@ -27,6 +30,16 @@ const { reducer, actions, name } = createSlice({
...
+      state.dataStatus = DataStatus.PENDING;
```
liza-veis: for authentification request let's use another property - authenticatedUserDataStatus

frontend/src/libs/components/app/app.tsx
```diff
@@ -16,6 +33,17 @@ const App: React.FC = () => {
...
+    if (!localStorage.getItem(StorageKey.TOKEN)) {
+      navigate(AppRoute.SIGN_IN);
+    }
+    void dispatch(authActions.getUser())
+      .unwrap()
+      .catch(() => {
+        navigate(AppRoute.SIGN_IN);
+      });
+  }, [dispatch, navigate]);
+
```
liza-veis, line 44: could you add a loader on the app when this status is not fulfilled?

liza-veis (changes requested):

frontend/src/libs/components/app/app.tsx
OlenaYatseniuk: I removed it in my issue (Protected Routing) pr #18, so you can just update the development branch for this purpose, I guess.

frontend/src/packages/auth/auth-api.ts
Nanaier: @Liza-Veis hey - but contentType is a mandatory field, so I can't delete it from the request - should I update the HTTPApiOptions type or leave it as it is?

```
type HTTPApiOptions = Omit<HTTPOptions, 'headers' | 'payload'> & {
  hasAuth: boolean;
  contentType: ValueOf<typeof ContentType>;
  payload?: HTTPOptions['payload'];
};
```

frontend/src/libs/components/app/app.tsx
```diff
@@ -16,27 +36,45 @@ const App: React.FC = () => {
...
+          </ul>
+          <p>Current path: {pathname}</p>
+
+          <div>
+            <RouterOutlet />
+          </div>
+        </>
+      ) : (
+        <Loader />
+      )}
```
HIMDanny: You also have the option to do like so:
```ts
if (authenticatedUserDataStatus !== DataStatus.FULFILLED) {
  return <Loader />;
}

return (
  <>
    <Header />
    <img src={reactLogo} className="App-logo" width="30" alt="logo" />

    <ul className="App-navigation-list">
      <li>
        <Link to={AppRoute.ROOT}>Root</Link>
      </li>
      <li>
        <Link to={AppRoute.SIGN_IN}>Sign in</Link>
      </li>
      <li>
        <Link to={AppRoute.SIGN_UP}>Sign up</Link>
      </li>
    </ul>
    <p>Current path: {pathname}</p>

    <div>
      <RouterOutlet />
    </div>
  </>
)
```

liza-veis: Yes, let's change to this

frontend/src/packages/auth/auth-api.ts
liza-veis: make this field optional please 🙏

frontend/src/libs/components/app/app.tsx
```diff
@@ -16,27 +36,45 @@ const App: React.FC = () => {
     }
   }, [isRoot, dispatch]);
 
+  useEffect(() => {
+    const checkTokenAndFetchUser = async (): Promise<void> => {
+      if (!(await storage.has(StorageKey.TOKEN))) {
+        navigate(AppRoute.SIGN_IN);
+      }
```
liza-veis: we should avoid side effects in components, this check should be in action, also we don't need to navigate to sign in, just do nothing in action if we have no token

navigation to sign-in should be handled by Protected Routes

frontend/src/slices/auth/actions.ts
```diff
@@ -18,4 +19,14 @@ const signUp = createAsyncThunk<
   return authApi.signUp(registerPayload);
 });
 
-export { signUp };
+const getAuthenticatedUser = createAsyncThunk<
+  UserAuthResponseDto,
+  undefined,
+  AsyncThunkConfig
+>(`${sliceName}/get-user`, (_, { extra }) => {
```
liza-veis:
```suggestion
>(`${sliceName}/get-authenticated-user`, (_, { extra }) => {
```

frontend/src/slices/auth/auth.slice.ts
```diff
@@ -2,14 +2,19 @@ import { createSlice } from '@reduxjs/toolkit';
...
+import { getAuthenticatedUser, signUp } from './actions.js';
 
 type State = {
+  user: UserGetAllItemResponseDto | null;
+  authenticatedUserDataStatus: ValueOf<typeof DataStatus>;
   dataStatus: ValueOf<typeof DataStatus>;
 };
 
 const initialState: State = {
+  user: null,
```
liza-veis: I was thinking about it, and let's rename user to authenticatedUser too, so we preserve consistency, and in case we have other users (like for forget password  we will need a temp user) it will be more obvious

liza-veis (approved):

liza-veis (changes requested):
Mixed up buttons 😅 Please fix comments above

shared/src/packages/auth/libs/enums/auth-api-path.enum.ts
```diff
@@ -1,6 +1,7 @@
 const AuthApiPath = {
   ROOT: '/',
   SIGN_UP: '/sign-up',
+  USER: '/authenticated-user',
```
what1s1ove: key and value should be the same

btw, usually this route is named as `/current`

frontend/src/packages/auth/auth-api.ts
```diff
@@ -35,6 +36,18 @@ class AuthApi extends BaseHttpApi {
...
+  public async getAuthenticatedUser(): Promise<UserAuthResponseDto> {
+    const response = await this.load(
+      this.getFullEndpoint(AuthApiPath.USER, {}),
+      {
+        method: 'GET',
+        hasAuth: true,
+      },
+    );
+
+    return await response.json<UserAuthResponseDto>();
```
v0ldemar01, line 49:
```suggestion
    return response.json<UserAuthResponseDto>();
```
The async function will wrap with Promise

frontend/src/libs/components/app/app.tsx
```diff
@@ -16,6 +33,14 @@ const App: React.FC = () => {
     }
   }, [isRoot, dispatch]);
 
+  useEffect(() => {
+    void dispatch(authActions.getAuthenticatedUser());
+  }, [dispatch]);
+
+  if (authenticatedUserDataStatus !== DataStatus.FULFILLED) {
+    return <Loader />;
+  }
```
fshabanov: I suggest we show the loader in Pending state, because currently the Loader will be displayed forever if the request gets rejected

frontend/src/packages/auth/auth-api.ts
```diff
@@ -35,6 +36,18 @@ class AuthApi extends BaseHttpApi {
 
     return await response.json<UserSignUpResponseDto>();
   }
+
+  public async getAuthenticatedUser(): Promise<UserAuthResponseDto> {
+    const response = await this.load(
+      this.getFullEndpoint(AuthApiPath.USER, {}),
```
v0ldemar01, line 42: Is it mandatory to provide empty object here?

v0ldemar01 (changes requested):

fshabanov (changes requested):

frontend/src/libs/components/app/app.tsx
liza-veis: Here I would prefer to leave it as it is

frontend/src/libs/components/app/app.tsx
```diff
@@ -16,6 +33,14 @@ const App: React.FC = () => {
     }
   }, [isRoot, dispatch]);
 
+  useEffect(() => {
+    void dispatch(authActions.getAuthenticatedUser());
+  }, [dispatch]);
+
+  if (authenticatedUserDataStatus !== DataStatus.FULFILLED) {
```
dmytro-kucherenko: As I understand, it will mount app component once (if I'm wrong, correct me, I'm just suggesting as what I see) and check if the authenticatedUserDataStatus is correct. Then it won't check it again, even if data status change, and therefore loader will stay. I think we should use: 

```
return (
   {authenticatedUserDataStatus !== DataStatus.FULFILLED ? <Loader /> : `other components`}
)
```

frontend/src/slices/auth/auth.slice.ts
```diff
@@ -2,14 +2,19 @@ import { createSlice } from '@reduxjs/toolkit';
...
 type State = {
+  authenticatedUser: UserGetAllItemResponseDto | null;
+  authenticatedUserDataStatus: ValueOf<typeof DataStatus>;
   dataStatus: ValueOf<typeof DataStatus>;
 };
 
 const initialState: State = {
+  authenticatedUser: null,
+  authenticatedUserDataStatus: DataStatus.IDLE,
   dataStatus: DataStatus.IDLE,
```
dmytro-kucherenko, line 18: Auth slice is responsible for authenticating the user and its data status is connected to it. I think there's no much sense in creating another new data status, if we already have one)

frontend/src/libs/components/app/app.tsx
liza-veis: @Dmytro-Kucherenko I like the comment you wrote, so keep going this way) Functional components works such a way that whenever your component re-renders the whole function (your component) will be executed again, that's why we need hooks that keep data despite re-renders. So these conditions above will work the same way)

frontend/src/slices/auth/auth.slice.ts
liza-veis: we usually keep authenticatedUser in auth state, because it's related to auth, of course it's not the only way, but in this project we will do so

frontend/src/libs/components/app/app.tsx
dmytro-kucherenko: Okay, thanks for the clarification)

shared/src/packages/auth/libs/enums/auth-api-path.enum.ts
liza-veis: please fix this comment) you should rename USER key to AUTHENTICATED_USER

liza-veis (commented):
Nice job 🙌  the last comment from me

liza-veis (dismissed):

fshabanov (dismissed):

v0ldemar01 (dismissed):

dmytro-kucherenko (dismissed):

OlesiaTr (commented):
Looks good to me as for now!

OlenaYatseniuk (dismissed):

frontend/src/libs/components/app/app.tsx
```diff
@@ -11,6 +22,12 @@ import styles from './styles.module.scss';
 const App: React.FC = () => {
   const { pathname } = useLocation();
   const dispatch = useAppDispatch();
+  const { authenticatedUserDataStatus } = useAppSelector(({ users, auth }) => ({
+    users: users.users,
+    dataStatus: users.dataStatus,
+    authenticatedUser: auth.authenticatedUser,
```
liza-veis:
```suggestion
```

liza-veis (changes requested):

liza-veis:
Convert this draft to usual pr

liza-veis (approved):

dmytro-kucherenko (approved):

BrianSammit (approved):

PR #47, opened 2023-08-22
cp-38: Add Bottom Tab Navigation

Vika030718:
- installed new packages:  @react-navigation/bottom-tabs, react-native-svg
- added a folder for icons 
- added a home svg icon
- implemented dummy home screen for testing bottom navigation
- implemented bottom tab navigation

This screen displays the bottom tab navigation
Screenshot 2023-08-21 at 23 30 16

This screen displays different states of the icons
Screenshot 2023-08-21 at 22 19 59


mobile/src/navigations/root/root.tsx
```diff
@@ -1,24 +1,56 @@
...
 
+const tabNavigatorOptions: BottomTabNavigationOptions = {
+  tabBarActiveTintColor: AppColor.BLUE_200,
+  tabBarIcon: HomeIcon,
+  tabBarInactiveTintColor: AppColor.GRAY_300,
+  tabBarShowLabel: false,
+  tabBarStyle: styles.tabBarStyle,
+};
+
+const BottomTabsNavigator: React.FC = () => {
```
fshabanov: I suggest we don't create all navigations in one file. Let's create a separate file for bottom tab navigation

fshabanov (changes requested):

mobile/src/navigations/root/root.tsx
Vika030718: Done, but I am not sure that I did it in the right way. Particularly, I am not sure about the naming conception - Main Stack, which has only TabNavigationBottom stack. Can it be done this way?

mobile/src/assets/img/icons/home-icon/home-icon.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
```
liza-veis: this should be in svg format, to change color dynamically you can set fill=currentColor, so path will be assets/img/icons/home.svg

mobile/src/libs/enums/navigation/root-screen-name.enum.ts
```diff
@@ -1,6 +1,8 @@
 const RootScreenName = {
-  SIGN_UP: 'Sign Up',
+  HOME: 'Home',
+  MAIN: 'Main Screens',
```
liza-veis: Home should be inside MainScreenName not RootScreenName
```suggestion
  MAIN: 'Main',
```

liza-veis (changes requested):

mobile/src/navigations/root/main/main.tsx
```diff
@@ -0,0 +1,32 @@
+import {
```
fshabanov: Move main/main.tsx out of root folder

fshabanov (changes requested):

mobile/src/navigations/root/root.tsx
fshabanov: Not really sure about MainStack here. At least because we are not using Stack navigator here. Maybe just Main?

mobile/src/screens/main/home.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+import { View } from '#libs/components/components';
+
+const Home: React.FC = () => {
+  return <View />;
```
danilittaa: I can suggest returning null instead of an empty `<View>`.This approach ensures that the component is ignored in the component tree, meaning an unnecessary View won't be created. However, I understand that this file might change later, so I don't think you need to make any changes now. Just a tip for the future)

iamAlinaaa (dismissed):
as for me code and component look good , i would advise you to delete unnecessary empty lines

mobile/src/navigations/main/main.tsx
```diff
@@ -0,0 +1,32 @@
...
+  type BottomTabNavigationOptions,
+  createBottomTabNavigator,
+} from '@react-navigation/bottom-tabs';
+import React from 'react';
+
+import { AppColor, MainScreenName } from '#libs/enums/enums';
+import { type TabNavigationParameterList } from '#libs/types/types';
+import { Home } from '#screens/main/home';
+
+import HomeIcon from '../../assets/img/icons/home.svg';
```
fshabanov:
```suggestion
import HomeIcon from '#assets/img/icons/home.svg';
```

fshabanov (changes requested):

mobile/src/screens/main/home.tsx
liza-veis: please fix this comment)

liza-veis (changes requested):
Good job! The last comment from me

liza-veis (dismissed):

mobile/src/navigations/main/main.tsx
Vika030718: done

mobile/src/navigations/root/root.tsx
```diff
@@ -8,6 +8,8 @@ import { RootScreenName } from '#libs/enums/enums';
 import { type RootNavigationParameterList } from '#libs/types/types';
 import { Auth } from '#screens/auth/auth';
 
+import { MainStack } from '../main/main';
```
fshabanov:
```suggestion
import { Main } from '../main/main';
```

fshabanov (dismissed):

OlesiaTr (dismissed):
Except written above issues - everything looks good to me! Good job!

mobile/src/navigations/root/root.tsx
Vika030718: Done

liza-veis (approved):

v0ldemar01 (approved):

Issue #16, opened 2023-08-19
Authorization token


liza-veis:
Implement JWT Token signing+verification

https://www.npmjs.com/package/jose

 - [ ] JWT token is generated and returned by Sign In / Sign Up
 - [x] JWT token stores userId
 - [x] JWT token is stored in localStorage

vladyslavdobychin:
High level implementation plan: 

- Locate Sign-in/Sign-up requests - how the request is sent from the front end. 
- Locate on the backend where the Sign-in/Sign-up requests are received and handled. 
- Add the logic to issue the JWT token after the successful verification.
- The verification process for the user should include the user ID, attach the user ID to JWT payload.
- The server needs to respond to the client after the verification, attach the token to the response body. 
- Locate on the client side where the response is received and save the token in the local storage

vladyslavdobychin:
Plan for updating the pull request: 

- [x] In the backend, in users.ts import config from #libs/packages/config/config.ts instead of creating a new instance 
- [x]  Remove comment from front end. 
- [x]  Create a separate JWT service. 
     - [x] Relocate the instance of JWT service to auth.ts 
     - [x] Update JWT service to receive the alg just like the secret key
     - [ ] Look into the option of turning the secret key and alg into their own variables and import them where needed
- [x] Add ALG as a parameter to the base-config.package.ts
- [x] Fix the CI issues with UserSignUpResponseDto not having a token
- [x] Remove build from included in the shared ts config
- [x] Make sure jose was removed from front-end 

TBD: 
- Clarify the necessity of including the build directory in the shared ts config 
> Was discussed, per Volodymyr Minchenko: That is the issue with eslint plugin and it won't influenced to linting the project. Typescript should check only src/ folder

vladyslavdobychin:
Previous changes were mostly accepted, adding new: 

- [x] Use Storage service instead of local storage on frontend 
- [x] Solve the problem with circular dependencies in the backend in users.ts 
>  [Per Vladimir](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/37#issuecomment-1689418394), the issue is cause by importing the user service to auth, which contains JWT service, and import JWT to user. As a solution, move JWT service to a separate module to **libs/packages/**

- [x] Figure out how to send token [separately](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/37#discussion_r1301862871), there should be a PR with it already. 
[...]

PR #48, opened 2023-08-22
cp-12: Add Sign In Flow

OlesiaTr:
- Implemented the sign-in user flow using the UserService class.
- Added a new route in the AuthController for user sign-in.
- Added the shared with FE validation schema to include validation rules for the sign-in request.
- Manually tested the new sign-in route using Postman.

cp-12-1
cp-12-2


backend/src/packages/users/user.repository.ts
```diff
@@ -42,6 +42,26 @@ class UserRepository implements Repository {
...
+
+    return UserEntity.initialize(user);
+  }
+
+  public async findById(id: number): Promise<UserEntity> {
+    const user = await this.userModel.query().findOne({ id }).execute();
+
+    if (!user) {
+      throw new Error('User not found');
+    }
```
v0ldemar01: Let's return null in this case

shared/src/packages/users/libs/validation-schemas/user-sign-in.validation-schema.ts
```diff
@@ -0,0 +1,24 @@
...
+        allow: false,
+      },
+    })
+    .required()
+    .messages({
+      'any.required': UserValidationMessage.EMAIL_REQUIRED,
+      'string.empty': UserValidationMessage.EMAIL_REQUIRED,
+      'string.email': UserValidationMessage.EMAIL_WRONG,
+    }),
+  password: joi.string().trim().required(),
```
v0ldemar01: Reuse validation schema from sign up, we have common components

v0ldemar01 (changes requested):

shared/src/packages/users/libs/validation-schemas/user-sign-in.validation-schema.ts
OlesiaTr: Should I create a reusable configuration for the email and password fields and reuse it in sign in/sign up schemas? And if yes, should they all be in separate files?

backend/src/packages/users/user.repository.ts
OlesiaTr: Done.

shared/src/packages/users/libs/validation-schemas/user-sign-in.validation-schema.ts
liza-veis: no, separate schemas for sign-up and sign-in, but they should be shared with backend

backend/src/packages/auth/auth.controller.ts
```diff
@@ -6,8 +6,14 @@ import {
...
-import { type UserSignUpRequestDto } from '#packages/users/users.js';
-import { userSignUpValidationSchema } from '#packages/users/users.js';
+import {
+  type UserSignInRequestDto,
+  type UserSignUpRequestDto,
+} from '#packages/users/users.js';
+import {
+  userSignInValidationSchema,
+  userSignUpValidationSchema,
+} from '#packages/users/users.js';
```
fshabanov:
```suggestion
import {
  type UserSignInRequestDto,
  type UserSignUpRequestDto,
  userSignInValidationSchema,
  userSignUpValidationSchema,
} from '#packages/users/users.js';
```

backend/src/packages/users/user.repository.ts
```diff
@@ -42,6 +52,34 @@ class UserRepository implements Repository {
   public delete(): ReturnType<Repository['delete']> {
     return Promise.resolve(true);
   }
+
+  public async findByEmail(email: string): Promise<UserEntity | null> {
+    const user = await this.userModel.query().findOne({ email }).execute();
```
fshabanov:
```suggestion
    const user = await this.userModel.query().findOne({ email });
```
I think we don't need it. Please, correct me if I'm wrong

fshabanov (changes requested):

backend/src/packages/users/user.entity.ts
```diff
@@ -2,87 +2,132 @@ import { type Entity } from '#libs/types/types.js';
...
+    return this.fullName;
+  }
+
+  public getCreatedAt(): Date | null {
+    return this.createdAt ?? null;
+  }
+
+  public getUpdatedAt(): Date | null {
+    return this.updatedAt ?? null;
+  }
```
HIMDanny: do we actually need this? We can call `toObject()` to get all these properties

backend/src/packages/users/user.repository.ts
```diff
@@ -16,7 +16,13 @@ class UserRepository implements Repository {
   public async findAll(): Promise<UserEntity[]> {
     const users = await this.userModel.query().execute();
 
-    return users.map((user) => UserEntity.initialize(user));
+    return users.map((user) =>
+      UserEntity.initialize({
+        ...user,
+        createdAt: new Date(user.createdAt),
```
HIMDanny, line 62: I've noticed you repeat this quite a few times. Maybe it is better to convert it into the date in the UserEntity constructor?

shared/src/packages/users/libs/validation-schemas/user-sign-in.validation-schema.ts
OlesiaTr: I am sorry, I am kind of confused. Should It be smth like this :

`const commonUserFields = {
  email: joi
    .string()
    .trim()
    .email({
      tlds: {
        allow: false,
      },
    })
    .required()
    .messages({
      'any.required': UserValidationMessage.EMAIL_REQUIRED,
      'string.empty': UserValidationMessage.EMAIL_REQUIRED,
      'string.email': UserValidationMessage.EMAIL_WRONG,
    }),
  password: joi.string().trim().required(),
};

const userSignUp = joi.object({
  ...commonUserFields,
  fullName: joi.string().trim().required().messages({
    'string.empty': UserValidationMessage.NAME_REQUIRED,
  }),
});

const userSignIn = joi.object({
  ...commonUserFields,
});` ?

backend/src/packages/auth/auth.controller.ts
```diff
@@ -75,6 +94,28 @@ class AuthController extends BaseController {
...
+    options: APIHandlerOptions<{
+      body: UserSignInRequestDto;
+    }>,
+  ): Promise<APIHandlerResponse> {
+    const user = await this.authService.verifyLoginCredentials(options.body);
+
+    if (!user) {
+      return {
+        status: HTTPCode.UNAUTHORIZED,
+        payload: { message: 'Unauthorized' },
```
dmytro-kucherenko: I think status should be NOT_FOUND with 404 code, because in fact verifyLoginCredentials checks, if the user with this data exists. Also the message will be transffered to frontend and as I understand will be displayed to the client, so it's better to make it more detailed, for instance 'User with these credentials was not found'

backend/src/packages/auth/auth.controller.ts
```diff
@@ -75,6 +94,28 @@ class AuthController extends BaseController {
...
+      };
+    }
+
+    await this.authService.signIn(user.id);
```
dmytro-kucherenko: May be I'm missing smth, but as I see you already got user above by email from userRepo. Therefore it's no reason in finding user again, because you already have it and it definitely will be in db

backend/src/packages/auth/auth.controller.ts
OlesiaTr: Done.

backend/src/packages/users/user.repository.ts
OlesiaTr: Changed.

backend/src/packages/users/user.entity.ts
OlesiaTr: Yes, you are right, forgot to get rid off it.

backend/src/packages/users/user.repository.ts
OlesiaTr: They have Date types in the User Entity already, but without it I get TS types conflict errors.

backend/src/packages/auth/auth.controller.ts
OlesiaTr: Since we are dealing with authentication and user login, if the user's credentials are invalid, returning an "Unauthorized" error makes more sense. This clearly communicates that the authentication process failed due to incorrect credentials, and the user should try again with valid credentials. And "Not Found" error is used when the requested resource is not found on the server. Question is then, at what point it is best to return "Unauthorized" - during password checking?

dmytro-kucherenko: In previous projects (for example backend structure) usually unauthorised was used in cases when token is wrong or client don't have enough permissions, while in login in case of not finding user with the same email was something like "User no found" exception

OlesiaTr: Ok, I added it!

backend/src/packages/auth/auth.controller.ts
OlesiaTr: I did it by example from the [mini-project](https://github.com/BinaryStudioAcademy/thread-js/blob/production/server/src/packages/auth/auth.service.js), if its not needed - I will change it.

shared/src/packages/users/libs/validation-schemas/user-sign-in.validation-schema.ts
OlesiaTr: Done.

backend/src/packages/auth/auth.controller.ts
dmytro-kucherenko: Oh, I see. In mini-project by making this step they're creating token and returning it with the user, so maybe it's supposed to be the same way here. Although pr of an issue #16 is already merged, I'm not sure who should implement jwt token in sign in/sign up, because this action is unchecked there

backend/src/packages/users/user.repository.ts
HIMDanny: I mean you can pass it like a string to the constructor and then convert to date
```ts
{
    this.id = id;
    this.email = email;
    this.fullName = fullName;
    this.passwordHash = passwordHash;
    this.passwordSalt = passwordSalt;
    this.createdAt = new Date(createdAt);
    this.updatedAt = new Date(updatedAt);
  }
```

HIMDanny: so that you can simply pass user object to the `initialize` method
```ts
UserEntity.initialize(user);
```

OlesiaTr: I will have a conflict with the User Model anyway, because it is based on the Abstract, that has createdAt and updatedAt as type string

OlesiaTr: @v0ldemar01 @fshabanov hello, would it be ok to change createdAt/updatedAt types in the Abstract model to Date instead of the string?

backend/src/packages/auth/auth.controller.ts
v0ldemar01: Great point, bcos as I see no token and jwt was used

backend/src/packages/users/user.repository.ts
v0ldemar01: Let's use @HIMDanny approach)

v0ldemar01: So converting string to Date will be in one place

backend/src/packages/users/user.service.ts
```diff
@@ -62,6 +64,20 @@ class UserService implements Service {
...
+  ): Promise<UserSignInResponseDto | null> {
+    const userEntity = await this.userRepository.findByEmail(email);
+
+    return userEntity ? userEntity.toObject() : null;
+  }
+
+  public async findById(id: number): Promise<UserSignInResponseDto | null> {
+    const userEntity = await this.userRepository.findById(id);
+
+    return userEntity ? userEntity.toObject() : null;
```
v0ldemar01:
```suggestion
    return userEntity?.toObject() ?? null;
```

v0ldemar01 (changes requested):

backend/src/packages/users/user.repository.ts
OlesiaTr: Like I said above, I have User Model conflicts because of Abstract model. So can I change it?

backend/src/packages/users/user.model.ts
```diff
@@ -10,6 +10,8 @@ class UserModel extends AbstractModel {
 
   public passwordSalt!: string;
 
+  public fullName!: string;
```
fshabanov: Full name should be inside user_details table, not users

shared/src/packages/users/libs/enums/user-validation-message.enum.ts
```diff
@@ -1,6 +1,7 @@
 const UserValidationMessage = {
   EMAIL_REQUIRED: 'Email is required',
   EMAIL_WRONG: 'Email is wrong',
+  NAME_REQUIRED: 'Name is required',
```
fshabanov: Please, put a dot at the end of each message

shared/src/packages/users/libs/validation-schemas/common-auth-fields.ts
```diff
@@ -0,0 +1,23 @@
+import joi from 'joi';
+
+import { UserValidationMessage } from '../enums/user-validation-message.enum.js';
+
+const commonUserFields = {
```
fshabanov: Name of the variable should be the same as the name of the file

fshabanov (changes requested):

backend/src/packages/users/user.repository.ts
v0ldemar01: So, userModel contains createdAt with type string. After response from database you will pass the result to `UserEntity.initialize`. `.initialize()` method will call UserEntity constructor with createdAt: string. The constructor reveives createdAt with string type and initializes t`his.createdAt = new Date(createdAt);`. So `UserEntity` createdAt will be with Date type and a constructor with string one

v0ldemar01 (changes requested):
Contact with sign up developer as for encrypt servvice

backend/src/packages/users/user.repository.ts
OlesiaTr: Finally done ;(

OlesiaTr: Thank you very much for your patience))

backend/src/packages/auth/auth.controller.ts
OlesiaTr: Should I delete signIn and findById methods too then, if they are not used atm?

backend/src/packages/users/user.model.ts
OlesiaTr: Deleted it.

shared/src/packages/users/libs/enums/user-validation-message.enum.ts
OlesiaTr: Added the dots.

shared/src/packages/users/libs/validation-schemas/common-auth-fields.ts
OlesiaTr: Changed the name of the variable.

liza-veis: Let's remove this and use separate schemas

liza-veis (changes requested):

shared/src/packages/users/libs/validation-schemas/common-auth-fields.ts
OlesiaTr: But I was told to reuse validation schema? Sign in/ sign up schemas are different, they just use this one cause of the common inputs. Or did I understood wrongly what changes I was supposed to do?

liza-veis: There are different approaches, but I would prefer to split them, because we will end up creating a lot of shared validation schemas, but when will need to have different validation, for the same field we will need either change a lot or override this field which will add complexity to the code. It's ok in some case, but for sign-up sign-in it's usual thing to have different validation, that's why it's better to split them

liza-veis: if you are referring to this comment https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/48#discussion_r1301949396, I mean that it should be in shared folder so we can reuse them on backend/frontend, but they should be split

backend/src/packages/users/user.entity.ts
```diff
@@ -2,49 +2,57 @@ import { type Entity } from '#libs/types/types.js';
...
     passwordSalt,
+    createdAt,
+    updatedAt,
   }: {
     id: number | null;
     email: string;
     passwordHash: string;
     passwordSalt: string;
+    createdAt: Date | string | null;
+    updatedAt: Date | string | null;
```
liza-veis: Hey) I know that there was a long discussion, but let's return back to the date approach here
```suggestion
    createdAt: Date | null;
    updatedAt: Date | null;
```
and map them when you call methods, as you did before, but with all properties, like this:
```
UserEntity.initialize({
  id: payload.id
  email: payload.email,
  fullName: payload.fullName,
  createdAt: new Date(payload.createdAt),
  updatedAt: new Date(payload.createdAt)
}),
```
sorry for that 🙏

liza-veis (changes requested):

HIMDanny (dismissed):
Look good to me. Just fix the issues listed above)

shared/src/packages/users/libs/validation-schemas/common-auth-fields.ts
```diff
@@ -0,0 +1,23 @@
+import joi from 'joi';
+
+import { UserValidationMessage } from '../enums/user-validation-message.enum.js';
```
fshabanov:
```suggestion
import { UserValidationMessage } from '../enums/enums.js';
```

shared/src/packages/users/libs/validation-schemas/common-auth-fields.ts
```diff
@@ -0,0 +1,23 @@
...
+        allow: false,
+      },
+    })
+    .required()
+    .messages({
+      'any.required': UserValidationMessage.EMAIL_REQUIRED,
+      'string.empty': UserValidationMessage.EMAIL_REQUIRED,
+      'string.email': UserValidationMessage.EMAIL_WRONG,
+    }),
+  password: joi.string().trim().required(),
```
fshabanov: Add messages for password as well

fshabanov (changes requested):

shared/src/packages/users/libs/validation-schemas/common-auth-fields.ts
OlesiaTr: Yes, ok))

backend/src/packages/users/user.entity.ts
OlesiaTr: It's ok, it happens. Thank you for clarifications))

backend/src/packages/auth/auth.controller.ts
```diff
@@ -75,6 +92,26 @@ class AuthController extends BaseController {
...
+    }>,
+  ): Promise<APIHandlerResponse> {
+    const user = await this.authService.verifyLoginCredentials(options.body);
+
+    if (!user) {
+      return {
+        status: HTTPCode.NOT_FOUND,
+        payload: { message: 'User with these credentials was not found' },
+      };
+    }
```
v0ldemar01: We need to handle it in service layer

v0ldemar01: So create a custom error for such purpose

v0ldemar01: Also don't forget to register it in error handling helper.

v0ldemar01: We already have that functionality)

v0ldemar01 (changes requested):

backend/src/packages/auth/auth.service.ts
```diff
@@ -16,6 +18,16 @@ class AuthService {
   ): Promise<UserSignUpResponseDto> {
     return this.userService.create(userRequestDto);
   }
+
+  public async verifyLoginCredentials({
+    email, // password, --> should be used for password comparison
```
v0ldemar01: After merging the Sign Up we need to add verification of password

v0ldemar01 (changes requested):

backend/src/packages/auth/auth.service.ts
OlesiaTr: Done.

OlenaYatseniuk (dismissed):

shared/src/libs/exceptions/incorrect-password-error/incorrect-password-error.exception.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { AuthError } from '../exceptions.js';
+
+type Constructor = {
+  message: string;
+  status?: ValueOf<typeof HTTPCode>;
+  cause?: unknown;
+};
+
+class IncorrectPasswordError extends AuthError {
```
OlesiaTr: Is it ok to use it or am I supposed to make InvalidCredentialsClass as it is in the mini-project?

Xxandder: I think you need to ask about specification for Sign In in the issue. On some websites you can see different errors: "user not found" or "incorrect password". But there is also websites where you can only see one error like "wrong email or password"

backend/src/libs/packages/server-application/libs/helpers/get-error-info/get-incorrect-password-error-info.helper.ts
Xxandder, line 1: Maybe I am mistaken, but you check if the password is correct, not the format. So I think here you shouln't use VALIDATION but maybe AUTHORIZATION one

backend/src/packages/auth/auth.controller.ts
```diff
@@ -90,6 +104,19 @@ class AuthController extends BaseController {
     };
   }
```
liza-veis, line 106: Add comments for swagger documentation

backend/src/packages/auth/auth.service.ts
```diff
@@ -18,6 +25,33 @@ class AuthService {
...
   }
 
+  public async verifyLoginCredentials({
+    email,
+    password,
+  }: UserSignInRequestDto): Promise<UserSignInResponseDto | null> {
+    const user = await this.userService.findByEmail(email);
+
+    if (!user) {
+      throw new UserNotFoundError({
```
liza-veis: We will use error per service, not per status or specific error. So instead of UserNotFoundError and IncorrectPasswordError, you should have UsersError and pass needed status code

backend/src/packages/auth/auth.service.ts
```diff
@@ -18,6 +25,33 @@ class AuthService {
...
+        message: 'User with these credentials was not found',
```
liza-veis: Use ExceptionMessage enum

backend/src/packages/auth/auth.service.ts
```diff
@@ -18,6 +25,33 @@ class AuthService {
...
+      });
+    }
+
+    const pswrdCompare = await encrypt.compare(password, user.passwordHash);
```
liza-veis: By quality criteria we should not use abbreviations
```suggestion
    const isSamePassword = await encrypt.compare(password, user.passwordHash);
```

shared/src/libs/packages/http/libs/enums/http-code.enum.ts
```diff
@@ -2,6 +2,8 @@ const HTTPCode = {
   OK: 200,
   CREATED: 201,
   UNAUTHORIZED: 401,
+  INCORRECT_PASSWORD: 401,
```
liza-veis: This enum is for standard http codes, not for mapping http codes to our functionality
```suggestion
```

liza-veis (changes requested):

backend/src/libs/packages/server-application/libs/helpers/get-error-info/get-incorrect-password-error-info.helper.ts
liza-veis: We will rename it to UsersError (as error per service), so it should be Users Error

dmytro-kucherenko (dismissed):

BrianSammit (dismissed):
With the changes above it should be good!! 😄

backend/src/packages/auth/auth.controller.ts
OlesiaTr: Done!

backend/src/packages/auth/auth.service.ts
OlesiaTr: Fixed it!

backend/src/packages/auth/auth.service.ts
OlesiaTr: I thought it was a common name, as in B16 examples. I changed it!

shared/src/libs/packages/http/libs/enums/http-code.enum.ts
OlesiaTr: Fixed it!

backend/src/packages/auth/auth.service.ts
OlesiaTr: Done!

backend/src/libs/packages/server-application/libs/helpers/get-error-info/get-incorrect-password-error-info.helper.ts
OlesiaTr: Yes, I used authorization instead, thank you!

backend/src/packages/auth/auth.service.ts
Xxandder, line 1: Maybe it's better to use status codes from HttpCode enum?

backend/src/libs/packages/server-application/libs/helpers/get-error-info/get-users-error-info.helper.ts
Xxandder, line 1: @Liza-Veis Hi, could you, please, explain why we use UsersErrorInfo instead of something like InvalidCredentialsError as in mini-project. I am just a bit confused with this and wanted to get the logic and in future use correct errors naming.

Xxandder (changes requested):

backend/src/libs/packages/server-application/libs/helpers/get-error-info/get-users-error-info.helper.ts
HIMDanny: hey! I think this comment will answer your question https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/48#discussion_r1306175317

liza-veis: Hey! in mini project we used InvalidCredentialsError for Authorization plugin and Auth service, so instead of it we have AuthError, and for user service we will have UsersError

backend/src/packages/auth/auth.service.ts
```diff
@@ -18,6 +23,35 @@ class AuthService {
...
+    password,
+  }: UserSignInRequestDto): Promise<UserSignInResponseDto | null> {
+    const user = await this.userService.findByEmail(email);
+
+    if (!user) {
+      throw new UsersError({
+        status: 404,
+        message: ExceptionMessage.USER_NOT_FOUND,
+      });
+    }
```
liza-veis: this should be thrown in users service when you call findByEmail, also let's user HttpCode.NOT_FOUND instead of number

backend/src/packages/auth/auth.service.ts
```diff
@@ -18,6 +23,35 @@ class AuthService {
...
+
+    const isSamePassword = await encrypt.compare(password, user.passwordHash);
+
+    if (!isSamePassword) {
+      throw new UsersError({
+        status: 401,
+        message: ExceptionMessage.INCORRECT_PASSWORD,
+      });
```
liza-veis: for errors related to auth we throw AuthError, also it's better to return incorrect credentials error message instead of pointing out what's wrong, this is for security purpose
```suggestion
      throw new AuthError({
        message: ExceptionMessage.INCORRECT_CREDENTIALS,
      });
```

backend/src/packages/users/user.service.ts
```diff
@@ -45,11 +45,7 @@ class UserService implements Service {
   ): Promise<ReturnType<UserEntity['toObject']> | null> {
     const user = await this.userRepository.findById(id);
 
-    if (!user) {
-      return null;
-    }
-
-    return user.toObject();
+    return user?.toObject() ?? null;
```
liza-veis, line 49: We should throw error here, instead of returning null

liza-veis (changes requested):

backend/src/packages/users/user.entity.ts
```diff
@@ -97,13 +97,15 @@ class UserEntity implements Entity {
...
     updatedAt: Date;
+    passwordHash: string;
   } {
     return {
       id: this.id as number,
       email: this.email,
       fullName: this.fullName,
       createdAt: this.createdAt as Date,
       updatedAt: this.updatedAt as Date,
+      passwordHash: this.passwordHash,
```
fshabanov: Be careful not to send this field to the frontend when we request users

shared/src/libs/exceptions/users-error/users-error.exception.ts
```diff
@@ -0,0 +1,25 @@
+import { type HTTPCode } from '#libs/packages/http/http.js';
+import { type ValueOf } from '#libs/types/value-of.type.js';
```
fshabanov:
```suggestion
import { type ValueOf } from '#libs/types/types.js';
```

fshabanov (changes requested):

backend/src/packages/users/user.entity.ts
liza-veis: yes, agree, let's remove passwordHash from toObject method and create a getter
```
  public get passwordData(): {
    passwordHash: string;
    passwordSalt: string;
  } {
    return {
      passwordSalt: this.passwordSalt as string,
      passwordHash: this.passwordHash as string,
    };
  }
```

v0ldemar01: We can just do it from model layer via
```js
public static override get modifiers(): Modifiers<QueryBuilder<User>> {
    return {
      withoutPassword(builder): QueryBuilder<User> {
        return builder.select('id', 'email', 'createdAt', 'updatedAt');
      },
    };
  }
```

v0ldemar01 (changes requested):

dmytro-kucherenko (dismissed):

backend/src/packages/auth/auth.service.ts
OlesiaTr: Fixed it!

backend/src/packages/auth/auth.service.ts
OlesiaTr: Fixed this!

backend/src/packages/users/user.service.ts
Xxandder: But if we throw error here we will never throw AuthError in case when user with such email doesn't exists. Or maybe I am wrong?

Xxandder: @Liza-Veis  I tried it in my code and it stops my app with thowen error even if I'm catching it

liza-veis: ok, got it, let's leave it then

backend/src/packages/users/user.entity.ts
OlesiaTr: I need passwordHash for password comparing, can I return user.toObject from auth.service that will be without passwordHash? Or can you clarify how I am supposed to use the code written by Volodymyr?

shared/src/libs/exceptions/users-error/users-error.exception.ts
OlesiaTr: Fixed.

backend/src/packages/users/user.entity.ts
v0ldemar01: Let's proceed with additional entity like UserWithPasswordEntity, so the default behaviour is to get user without password fields

Nanaier (dismissed):

shared/src/libs/enums/exception-message.enum.ts
```diff
@@ -1,6 +1,8 @@
 const ExceptionMessage = {
   INVALID_TOKEN: 'Token is invalid.',
   UNAUTHORIZED_USER: 'User is not authorized.',
+  USER_NOT_FOUND: 'User with these credentials was not found',
```
liza-veis: End messages with period

backend/src/packages/users/user.model.ts
```diff
@@ -22,6 +27,20 @@ class UserModel extends AbstractModel {
     return DatabaseTableName.USERS;
   }
 
+  public static override get modifiers(): Modifiers<QueryBuilder<UserModel>> {
+    return {
+      withoutPassword(builder): QueryBuilder<UserModel> {
```
v0ldemar01, line 32: So it's only declaration but never used.

backend/src/packages/users/user-with-password.entity.ts
```diff
@@ -0,0 +1,141 @@
...
+      fullName,
+      createdAt,
+      updatedAt,
+      isSurveyCompleted,
+    });
+    this.passwordHash = passwordHash;
+    this.passwordSalt = passwordSalt;
+  }
+
+  public getPasswordInfo(): {
```
v0ldemar01: Is it used anythere?

v0ldemar01 (changes requested):

backend/src/packages/users/user-with-password.entity.ts
```diff
@@ -0,0 +1,141 @@
...
+    passwordHash: string;
+    passwordSalt: string;
+  } {
+    return {
+      passwordHash: this.passwordHash,
+      passwordSalt: this.passwordSalt,
+    };
+  }
+
+  public static initializeWithPassword({
```
liza-veis: let's override all methods and name them as in extended model
```suggestion
  public static initialize({
```

backend/src/packages/users/user.repository.ts
```diff
@@ -132,6 +113,54 @@ class UserRepository implements Repository {
...
+      email: user.email,
+      passwordHash: user.passwordHash,
+      passwordSalt: user.passwordSalt,
+      createdAt: new Date(user.createdAt),
+      updatedAt: new Date(user.updatedAt),
+      fullName: user.details?.fullName ?? '',
+      isSurveyCompleted: user.details?.isSurveyCompleted ?? false,
+    });
+  }
+
```
liza-veis: We added findByEmail in another pr, you will see it after you resolve merge conflicts. We don't need to do anything with it, just remove passwordHash and passwordSalt from it and from UserCommonQueryResponse.

For getting with password we will have separate method (as it's done now), that won't call getByEmail and will repeat the logic a bit, except using different types. We will need to create another type UserWithDetailsQueryResponse that will extends UserCommonQueryResponse and add passwordHash and passwordSalt to it

```suggestion
  public async findByEmailWithPassword(
    email: string,
  ): Promise<UserWithPasswordEntity | null> {
    const user = await this.userModel
      .query()
      .withGraphJoined(UsersRelation.DETAILS)
      .findOne({ email })
      .castTo<UserWithPasswordQueryResponse | undefined>();

    if (!user) {
      return null;
    }

    return UserWithPasswordEntity.initializeWithPassword({
      id: user.id,
      email: user.email,
      passwordHash: user.passwordHash,
      passwordSalt: user.passwordSalt,
      createdAt: new Date(user.createdAt),
      updatedAt: new Date(user.updatedAt),
      fullName: user.details?.fullName ?? '',
      isSurveyCompleted: user.details?.isSurveyCompleted ?? false,
    });
  }

```

backend/src/libs/packages/server-application/libs/helpers/get-error-info/get-users-error-info.helper.ts
```diff
@@ -0,0 +1,20 @@
...
+const getUsersErrorInfo = (error: UsersError): ErrorInfo => {
+  const { message } = error;
+  const status = error.status;
+
+  return {
+    status,
+    internalMessage: `[Auth Error]: ${status} — ${message}`,
+    response: {
+      message,
+      errorType: ServerErrorType.AUTHORIZATION,
```
liza-veis:
```suggestion
    internalMessage: `[Users Error]: ${status} — ${message}`,
    response: {
      message,
      errorType: ServerErrorType.USERS,
```

backend/src/packages/auth/auth.service.ts
```diff
@@ -31,6 +34,29 @@ class AuthService {
...
   }
 
+  public async verifyLoginCredentials({
+    email,
+    password,
+  }: UserSignInRequestDto): Promise<UserSignInResponseDto | null> {
+    const user = await this.userService.findByEmail(email);
+    const userWithPassword = await this.userService.getUserInfoWithPassword(
+      email,
+    );
```
liza-veis: we need only one query, and then call UserEntity.initialize with it
```suggestion
    const user = await this.userService.getUserInfoWithPassword(
      email,
    );
```

liza-veis (changes requested):

backend/src/packages/users/user-with-password.entity.ts
liza-veis: please remove this method

backend/src/packages/auth/auth.service.ts
```diff
@@ -21,7 +24,7 @@ class AuthService {
     const { email } = userRequestDto;
 
     const user = await this.userService.findByEmail(email);
-    if (user) {
+    if (user.id && user.email) {
```
liza-veis:
```suggestion
    if (user) {
```

liza-veis (changes requested):

backend/src/packages/users/user.model.ts
liza-veis: you need to add this modifier (withoutPassword) everywhere where you execute query and expect user without password  https://vincit.github.io/objection.js/recipes/modifiers.html

backend/src/packages/users/user-with-password.entity.ts
OlesiaTr: yes, sorry, wrote it to see the logic on how to use it and forgot to remove.

shared/src/libs/enums/exception-message.enum.ts
OlesiaTr: Done.

backend/src/packages/users/user.repository.ts
OlesiaTr: Yes, done!

backend/src/libs/packages/server-application/libs/helpers/get-error-info/get-users-error-info.helper.ts
OlesiaTr: Yes, fixed it!

backend/src/packages/auth/auth.service.ts
OlesiaTr: I can not commit with such conditional check
user-error

OlesiaTr: Thats why I had to change it

fshabanov: Linter is actually right, you will always have a user there. findByEmail will throw an error if there's no user, that means in that line in auth service we will always have a user and that will break the signUp functionality. I suggest in user service we return null if there is no user, instead of throwing an error

backend/src/packages/users/user.model.ts
OlesiaTr: Yes, that's what I was asking about [in another thread](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/48#discussion_r1307210242), am I supposed to use it with additional entity or not now?

liza-veis: you should use it where UserEntity is used, because it is without password, and you should not use it where we need UserWithPasswordEntity

backend/src/packages/auth/auth.service.ts
OlesiaTr: Yes, changed it!

backend/src/packages/auth/auth.service.ts
OlesiaTr: Changed it!

backend/src/packages/users/user-with-password.entity.ts
OlesiaTr: Done.

backend/src/packages/users/user.model.ts
OlesiaTr: I added it.

shared/src/libs/types/server-error-response.type.ts
```diff
@@ -1,5 +1,4 @@
-import { type ServerErrorType } from '#libs/enums/enums.js';
-
+import { type ServerErrorType } from '../../libs/enums/enums.js';
```
liza-veis:
```suggestion
import { type ServerErrorType } from '#libs/enums/enums.js';
```

backend/src/packages/users/user.service.ts
```diff
@@ -98,6 +102,33 @@ class UserService implements Service {
...
+  public async findByEmailWithPassword(
+    email: string,
+  ): Promise<ReturnType<UserWithPasswordEntity['toObject']>> {
+    const userEntity = await this.userRepository.findByEmailWithPassword(email);
+
+    if (!userEntity) {
+      throw new UsersError({
+        status: HTTPCode.NOT_FOUND,
+        message: ExceptionMessage.USER_NOT_FOUND,
+      });
```
liza-veis: let's return null here and throw this error in AuthService where we call this method (getByEmailWithPassword)

liza-veis (changes requested):

shared/src/libs/types/server-error-response.type.ts
OlesiaTr: Done!

backend/src/packages/users/user.service.ts
OlesiaTr: Done!

liza-veis (approved):

v0ldemar01 (approved):

PR #46, opened 2023-08-22
cp-14: Add sign up functionality

Xxandder:
- Created migration file that creates user_details table
- Created model for user_details
- Set up relation between users and user_details table
- Implemented saving registration data
- Implelented passwordHash generation
- Added new fields to userRequestDto and userResponseDto and files that use these types
image

PR #58, opened 2023-08-23
cp-56: Add chats and chat messages tables

ossavemu:
I have added the tables that chats and chat messages, creating the corresponding migration files. I created a helper to avoid repeating code and updated the default migration to use the helper.

image

image


backend/src/db/migrations/20230215152530_add_users_table.ts
Xxandder, line 1: Hi, I think here you need to get rid of TableNames and UserTableColumns enums.  I had the same problem : [here](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/46#discussion_r1301949765)
And the same for other another migration

backend/src/db/migrations/20230822123000_add_chat_messages_table.ts
Xxandder, line 1: You also need to add onDelete and onUpdate for foreign columns so messages will be automaticly deleted/updated when a chat or sender is deleted/updated

Xxandder: Sorry for missing changes blocks in my comments. I just didn't know about how to add it

backend/src/db/migrations/20230215152530_add_users_table.ts
liza-veis: Agree

backend/src/db/helpers/migration-helper.ts
```diff
@@ -0,0 +1,34 @@
+import { type Knex } from 'knex';
+
+import { CommonTableColumns } from '#libs/enums/enums.js';
+
+function createTableWithCommonColumns(
```
v0ldemar01: As @Xxandder mentioned, we need to create an independent migrations. So your idea is good, bit not suitable for migrations

backend/src/db/migrations/20230215152530_add_users_table.ts
```diff
@@ -1,35 +1,20 @@
...
 
-const ColumnName = {
-  ID: 'id',
-  EMAIL: 'email',
-  PASSWORD_HASH: 'password_hash',
-  PASSWORD_SALT: 'password_salt',
-  CREATED_AT: 'created_at',
-  UPDATED_AT: 'updated_at',
-};
+const up = createTableWithCommonColumns(
```
v0ldemar01: It is not a good idea to modify existing migrations

backend/src/libs/enums/chat-messages-table-columns.ts
```diff
@@ -0,0 +1,8 @@
+const ChatMessagesTableColumns = {
+  NAME: 'name',
+  CHAT_ID: 'chat_id',
+  SENDER_ID: 'sender_id',
+  MESSAGE: 'message',
+};
```
v0ldemar01:
```suggestion
} as const;
```

backend/src/libs/enums/chats-table-columns.ts
```diff
@@ -0,0 +1,5 @@
+const ChatsTableColumns = {
+  NAME: 'name',
+};
```
v0ldemar01:
```suggestion
} as const;
```

backend/src/libs/enums/common-table-columns.ts
```diff
@@ -0,0 +1,7 @@
+const CommonTableColumns = {
+  ID: 'id',
+  CREATED_AT: 'created_at',
+  UPDATED_AT: 'updated_at',
+};
```
v0ldemar01:
```suggestion
} as const;
```

backend/src/libs/enums/chats-table-columns.ts
```diff
@@ -0,0 +1,5 @@
+const ChatsTableColumns = {
```
v0ldemar01: Enums should be in singular form

backend/src/libs/enums/chat-messages-table-columns.ts
```diff
@@ -0,0 +1,8 @@
+const ChatMessagesTableColumns = {
```
v0ldemar01: Enums should be in singular form

backend/src/libs/enums/table-names.ts
```diff
@@ -0,0 +1,7 @@
+const TableNames = {
```
v0ldemar01: We already have DatabaseTableName

backend/src/libs/enums/user-table-columns.ts
```diff
@@ -0,0 +1,7 @@
+const UserTableColumns = {
```
v0ldemar01: It's not a correct destination for that enum

v0ldemar01: And please contact with guys which are have the tasks related to user table columns, maybe they are done with it

v0ldemar01 (changes requested):

backend/src/libs/enums/table-names.ts
ossavemu: I used the cp-14: Adding registry functionality #46 as an example when adding the enum

backend/src/libs/enums/database-table-name.enum.ts
v0ldemar01, line 1: image

backend/src/libs/enums/table-columns/chat-messages-table-column.enum.ts
```diff
@@ -0,0 +1,10 @@
+const ChatMessagesTableColumn = {
```
v0ldemar01: Please move to appropriate package

v0ldemar01 (changes requested):

liza-veis (approved):
Good job 🙌

Nanaier (approved):

fshabanov (approved):

v0ldemar01 (approved):

liza-veis (approved):

PR #63, opened 2023-08-23
cp-42: Add Sidebar

Nanaier:
Added the sidebar component. Multiple icons are shown for testing purposes + they redirect to different pages.
Screenshot 2023-08-23 at 21 30 10


frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,71 @@
+import { useLocation } from 'react-router-dom';
```
v0ldemar01: Move it to hooks

frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,71 @@
...
+          <div className={styles['image']}>
+            <img src={logo} alt="logo" />
+          </div>
+        </Link>
+      </nav>
+      <nav className={styles['nav']}>
+        <div className={styles['icons-container']}>
+          <button
+            className={`${
+              pathname === AppRoute.ROOT
```
v0ldemar01: Let's use getValidClassNames and iconSelected as an optional while condition is true

v0ldemar01 (changes requested):

BrianSammit (dismissed):

iamAlinaaa (dismissed):

frontend/src/assets/css/scaffolding.scss
```diff
@@ -6,6 +6,7 @@ body {
   display: flex;
   flex-direction: column;
   min-width: 320px;
+  height: 100vh;
```
liza-veis: this will cause issues with keyboard on android devices if we will have inputs on a page 🥲  this can be solved with height 100% either here, or maybe if we add it to root
```suggestion
```

frontend/src/assets/css/variables.scss
```diff
@@ -1,6 +1,7 @@
 :root {
   --white: #ffffff;
   --shadow: 0 2px 4px #0000001a;
+  --sidebar-shadow: 0px 4px 23px 0px #e5e1ff;
```
liza-veis: Let's not create a variable for this, this is to specific

liza-veis (changes requested):

frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,68 @@
+import home from '#assets/img/home.svg';
+import logo from '#assets/img/logo.svg';
+import { Link } from '#libs/components/components.js';
+import { AppRoute } from '#libs/enums/enums.js';
+import { getValidClassNames } from '#libs/helpers/helpers.js';
+import { useLocation } from '#libs/hooks/hooks.js';
+
+import styles from './styles.module.scss';
+
+const Sidebar: React.FC = () => {
```
liza-veis: It's better to pass config here and generate items according to it

frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,68 @@
...
+          </button>
+          <button
+            className={getValidClassNames(
+              styles[iconStyles],
+              pathname === AppRoute.SIGN_UP && styles[iconSelected],
+            )}
+          >
+            <Link to={AppRoute.SIGN_UP}>
+              <div>
+                <img src={home} alt="home" className={styles['icon']} />
```
liza-veis: Check A2 criteria, we need a label here with visibility-hidden class, we already have it in global styles

frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,68 @@
...
+            </Link>
+          </button>
+          <button
+            className={getValidClassNames(
+              styles[iconStyles],
+              pathname === AppRoute.SIGN_UP && styles[iconSelected],
+            )}
+          >
+            <Link to={AppRoute.SIGN_UP}>
+              <div>
```
liza-veis: let's use span instead of div here, not to nest block element inside inline element

liza-veis (changes requested):

frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,68 @@
...
+              styles[iconStyles],
+              pathname === AppRoute.ROOT && styles[iconSelected],
+            )}
+          >
+            <Link to={AppRoute.ROOT}>
+              <div>
+                <img src={home} alt="home" className={styles['icon']} />
+              </div>
+            </Link>
+          </button>
```
HIMDanny: Maybe we can use navlink here as it has an active class? So then there would be no need to check whether it is a valid path

OlesiaTr (dismissed):
Great job, after implementing comments above, I think its good to go!

HIMDanny (dismissed):
Good job!

frontend/src/libs/components/sidebar/sidebar.tsx
liza-veis: Thank you for the comment) in react native navigation works a bit differently, so we don't have NavLink

frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,68 @@
...
+            <img src={logo} alt="logo" />
+          </div>
+        </Link>
+      </nav>
+      <nav className={styles['nav']}>
+        <div className={styles['icons-container']}>
+          <button
+            className={getValidClassNames(
+              styles[iconStyles],
+              pathname === AppRoute.ROOT && styles[iconSelected],
```
ossavemu: This can be stored in a constant: `const iconSelectedStyles =  if(pathname === AppRoute.ROOT)  styles[iconSelected]; `

frontend/src/libs/components/sidebar/styles.module.scss
```diff
@@ -0,0 +1,66 @@
...
+.icon-selected {
+  position: relative;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  padding: 14px;
+  background-color: var(--white);
+  border: 0;
+  border-radius: 50%;
+  box-shadow: 0 4px 23px 0 #e5e1ff;
```
ossavemu: I think the color #e5e1ff can be saved as violet-100 css variable

frontend/src/assets/css/variables.scss
Nanaier: @Liza-Veis, hey I checked the Figma design and it seems that we are going to use these specific colors(`0px 4px 23px 0px #e5e1ff`) for a lot of shadows in the future - should I save the #e5e1ff as a global color - if so would violet-100 work?

frontend/src/libs/components/sidebar/sidebar.tsx
Nanaier: I added the changes here, but not entirely sure it is what you wanted

frontend/src/assets/css/variables.scss
liza-veis: Yes

frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { getValidClassNames } from '#libs/helpers/helpers.js';
+import { useLocation } from '#libs/hooks/hooks.js';
+import { type ValueOf } from '#libs/types/types.js';
+
+import styles from './styles.module.scss';
+
+type Properties = {
+  routes: {
+    route: ValueOf<typeof AppRoute>;
+    routeName: string;
```
liza-veis:
```suggestion
    path: ValueOf<typeof AppRoute>;
    name: string;
```

frontend/src/libs/components/sidebar/sidebar.tsx
```diff
@@ -0,0 +1,64 @@
...
+    icon: string;
+  }[];
+};
+
+const Sidebar: React.FC<Properties> = ({ routes }) => {
+  const iconSelected = 'icon-selected';
+  const iconStyles = 'icon-container';
```
liza-veis: Let's use them without variables
```suggestion
```

liza-veis (changes requested):

liza-veis (approved):

v0ldemar01 (approved):

PR #74, opened 2023-08-25
cp-29: Authenticated user flow

HIMDanny:
- Created a method for getting current user in the auth service
- Extended handler options with the authorized user object
- Created an endpoint for getting the current user in the auth controller

### DIFF
--- a/backend/src/libs/packages/controller/libs/types/api-handler-options.type.ts
+++ b/backend/src/libs/packages/controller/libs/types/api-handler-options.type.ts
@@ -4,7 +4,7 @@ type DefaultApiHandlerOptions = {
   body?: unknown;
   query?: unknown;
   params?: unknown;
-  user?: UserAuthResponseDto | null;
+  user?: UserAuthResponseDto;
 };
 
 type APIHandlerOptions<
--- a/backend/src/libs/packages/controller/libs/types/api-handler-options.type.ts
+++ b/backend/src/libs/packages/controller/libs/types/api-handler-options.type.ts
@@ -2,6 +2,7 @@ type DefaultApiHandlerOptions = {
   body?: unknown;
   query?: unknown;
   params?: unknown;
+  user?: unknown;
 };
 
 type APIHandlerOptions<
@@ -10,6 +11,7 @@ type APIHandlerOptions<
   body: T['body'];
   query: T['query'];
   params: T['params'];
+  user: T['user'];
 };
 
 export { type APIHandlerOptions };
--- a/backend/src/libs/packages/controller/libs/types/api-handler-options.type.ts
+++ b/backend/src/libs/packages/controller/libs/types/api-handler-options.type.ts
@@ -1,8 +1,10 @@
+import { type UserAuthResponseDto } from '#packages/users/users.js';
+
 type DefaultApiHandlerOptions = {
   body?: unknown;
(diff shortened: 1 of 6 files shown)


backend/src/libs/packages/controller/libs/types/api-handler-options.type.ts
```diff
@@ -2,6 +2,7 @@ type DefaultApiHandlerOptions = {
   body?: unknown;
   query?: unknown;
   params?: unknown;
+  user?: unknown;
```
v0ldemar01: let's provide type for it, auth user can have only single one

v0ldemar01 (changes requested):

backend/src/packages/auth/auth.controller.ts
```diff
@@ -33,6 +36,17 @@ class AuthController extends BaseController {
           }>,
         ),
     });
+
```
liza-veis, line 39: Add comments for swagger documentation as it's done for user service

backend/src/packages/auth/auth.controller.ts
```diff
@@ -75,6 +89,15 @@ class AuthController extends BaseController {
       payload: await this.authService.signUp(options.body),
     };
   }
+
+  private async getUser(
```
liza-veis:
```suggestion
  private async getAuthenticatedUser(
```

backend/src/packages/auth/auth.service.ts
```diff
@@ -16,6 +17,10 @@ class AuthService {
   ): Promise<UserSignUpResponseDto> {
     return this.userService.create(userRequestDto);
   }
+
+  public getUser(id: number): Promise<UserAuthResponseDto | null> {
```
liza-veis:
```suggestion
  public getAuthenticatedUser(id: number): Promise<UserAuthResponseDto | null> {
```

liza-veis (changes requested):

backend/src/packages/auth/auth.controller.ts
HIMDanny: done. Please tell me if something should be changed

backend/src/libs/packages/controller/libs/types/api-handler-options.type.ts
```diff
@@ -1,7 +1,10 @@
+import { type UserAuthResponseDto } from '#packages/users/users.js';
+
 type DefaultApiHandlerOptions = {
   body?: unknown;
   query?: unknown;
   params?: unknown;
+  user?: UserAuthResponseDto | null;
```
v0ldemar01: Not sure that user will have null value

v0ldemar01 (changes requested):

backend/src/libs/packages/controller/libs/types/api-handler-options.type.ts
HIMDanny: so then I should change a user param to optional in the fastify request as well

v0ldemar01: Good point, got it

v0ldemar01 (approved):

fshabanov (approved):

liza-veis (approved):

PR #79, opened 2023-08-25
cp-72: Swagger documentation is not public

HIMDanny:
- Created a served page path constant
- Added all unregistered routes to the whitelist


liza-veis (dismissed):

iamAlinaaa (dismissed):

backend/src/libs/packages/server-application/libs/constants/api.constants.ts
```diff
@@ -1,9 +1,12 @@
...
+const SERVED_PAGE_PATH = '/*';
+
 const WHITE_ROUTES = [
   { path: `/api/v1${APIPath.AUTH}${AuthApiPath.SIGN_UP}`, methods: ['POST'] },
   { path: `/api/v1${APIPath.AUTH}${AuthApiPath.SIGN_IN}`, methods: ['POST'] },
+  { path: `${SERVED_PAGE_PATH}`, methods: ['GET'] },
 ] as const;
 
-export { WHITE_ROUTES };
+export { SERVED_PAGE_PATH, WHITE_ROUTES };
```
fshabanov: Do we need to export it?

fshabanov (dismissed):

backend/src/libs/packages/server-application/libs/constants/api.constants.ts
HIMDanny: not sure) Will remove it from export

fshabanov (dismissed):

liza-veis (approved):

dmytro-kucherenko (dismissed):

BrianSammit (dismissed):

OlenaYatseniuk (dismissed):

fshabanov:
I tested this solution locally and it did not work. The problem is that 'checkIsWhiteRoute' helper does not check the patterns, it checks the exact routes. Were you able to open swagger documentation with this fix?

HIMDanny:
> I tested this solution locally and it did not work. The problem is that 'checkIsWhiteRoute' helper does not check the patterns, it checks the exact routes. Were you able to open swagger documentation with this fix?

The problem is that the documentation route is not recognized with `request.routerPath` and the documentation route is shown as `/*`. Furthermore, when accessing the documentation route after the fix, it returns the not found error. I guess it is a problem with swagger documentation

backend/src/libs/packages/server-application/libs/constants/api.constants.ts
```diff
@@ -1,9 +1,12 @@
 import { APIPath } from '#libs/enums/enums.js';
 import { AuthApiPath } from '#packages/auth/auth.js';
 
+const SERVED_PAGE_PATH = '/*';
+
 const WHITE_ROUTES = [
   { path: `/api/v1${APIPath.AUTH}${AuthApiPath.SIGN_UP}`, methods: ['POST'] },
   { path: `/api/v1${APIPath.AUTH}${AuthApiPath.SIGN_IN}`, methods: ['POST'] },
```
liza-veis: Let's do it this way:
1. Add /documentation to APIPath enum
2. In initMiddlewares replace routePrefix for swaggerUi with `/api/${api.version}/${APIPath.DOCUMENTATION}`.
This step will make it accessible from frontend by /api/v1/documentation route
3. in WHITE_ROUTES wrap paths in new RegExp. Don't forget to add ^ - start,  $ - end, so that we match exact routes for sign_up / sign_in
4. for SERVED_PAGE_PATH add \ before *, so that we match exact route, not any path
5. Add new white-route - ```new RegExp(`/api/v1${APIPath.DOCUMENTATION}/*`)```
6. in checkIsWhiteRoute helper use test instead of explicit comparison

After this it should be accessible from backend as well ✨ 

Note: don't forget to test RegExps that they match correct routes

liza-veis (changes requested):

backend/src/libs/packages/server-application/libs/constants/api.constants.ts
HIMDanny: can't we simply change `request.routerPath` to `request.url`? With this change, it understands the `api/v1/documentaion` route

HIMDanny: do you have other response then `not found` without the authorization plugin?

liza-veis: we will still need to allow served pages for the deploy because backend will need to return static pages

HIMDanny: but with this fix it is allowed

HIMDanny: now you can access all unregistered routes

HIMDanny: you can disconnect the authorization plugin and you'll see, that nothing actually changed. The documentation route will not be found
image

liza-veis: just checked, this won't work if we return static pages from backend (needed for deploy). Because instead of /* it will return /sign-in for example and we won't be able to detect that it's static except if we check that route doesn't start with /api/v1, but this approach is not good

HIMDanny: so I have to do the suggested above method anyway, right?

liza-veis: yep

HIMDanny: ok, no problem then)

HIMDanny: @Liza-Veis, could you help me understand, what you mean by `Because instead of /* it will return /sign-in for example and we won't be able to detect that it's static`? When I tried to access a `sign-in` route it showed me `/*`. Or there would be another response in production? Just want to clarify for myself)
image

liza-veis: Simple steps to reproduce how deploy will work:
1. run `npm run build -w frontend`
2. copy frontend build folder to backend folder and rename it to public
3. run  `npm run start-dev -w backend`
4. go to `localhost:3001/sign-in`
5. it should return sign-in page
6. remove public folder from backend (for cleanup after you check everything)

liza-veis: UPD: even with your steps above, you can see that request.url returns /sign-up instead of /*, so if we change request.routerPath to request.url, it won't work

HIMDanny: Should something be wrong while reproducing? Because I am receiving the `sign-in` page on frontend

HIMDanny: ok, I'll just make the suggested changes, cause I think we can discuss for a long time)

liza-veis: hmm that's strange because I'm getting /sign-up in request.url

HIMDanny: image

HIMDanny: sorry for confusing, I am getting this in request.routePath

HIMDanny: the only problem is that I am getting a not found from frontend part, when trying to access the `/api/v1/documentation` route

HIMDanny: I will try your suggested way and we'll see

liza-veis (approved):

backend/src/libs/packages/server-application/libs/constants/api.constants.ts
```diff
@@ -1,9 +1,19 @@
 import { APIPath } from '#libs/enums/enums.js';
 import { AuthApiPath } from '#packages/auth/auth.js';
 
+const SERVED_PAGE_PATH = '/\\*';
```
liza-veis: One note
```suggestion
const SERVED_PAGE_PATH = '^/\\*$';
```

liza-veis (changes requested):

liza-veis (approved):

fshabanov (approved):

PR #83, opened 2023-08-26
cp-76: Add Splash Screen

danilittaa:
- Added splash screen
- Automatically dissappers when application is loaded
- There is no design for this screen, so I improvised a little)

image


mobile/src/libs/components/app/app.tsx
```diff
@@ -11,6 +12,8 @@ import { Root as RootNavigation } from '#navigations/navigations';
 import { styles } from './styles';
 
 const App: React.FC = () => {
+  SplashScreen.hide();
```
liza-veis: all side effects should be in useEffect

liza-veis (changes requested):
I like the design

liza-veis (approved):

fshabanov (approved):

v0ldemar01 (approved):

iamAlinaaa (approved):
Looks great 🤩

PR #66, opened 2023-08-24
cp-50: Add survey endpoint and serveys db

OlenaYatseniuk:
- Add endpoint to save survey /sign-up/survey (create survey controller, service, entity, model, repository)
- Add a new migration file for creating a survey table in the database
- Add a new migration file for  adding the is_survey_completed flag to the user's table ( also update user entity)
- Add updateIsSurveyCompleted methods in the user repository and service to receive this flag when getting users 
- Update database schema in readme
photo_5422717926181556091_y
photo_5422717926181556098_w

PR #88, opened 2023-08-27
cp-70: Add file uploading functionality

Nanaier:
Added AWS file uploading functionality: 
- Created migrations for files table
- Created model, entity, repository, controller, service for files 
- Updated env with new credentials


ossavemu:
You could add a description, or a screenshot to the Pull Request. According to the code, I noticed that you have not been passed the AWS credentials, how did you test that?

backend/src/libs/plugins/file-uploading/file-uploading.plugin.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import { ExceptionMessage } from '#libs/enums/enums.js';
+import { FileError } from '#libs/exceptions/exceptions.js';
+import { ControllerHook } from '#libs/packages/controller/controller.js';
+
+type Options = {
+  extensions: string[];
+};
+
+const fileUploading = fp<Options>((fastify, { extensions }, done) => {
```
v0ldemar01:
```suggestion
const fileUpload = fp<Options>((fastify, { extensions }, done) => {
```

backend/src/packages/files/file.service.ts
```diff
@@ -0,0 +1,123 @@
...
+    const fileKey = `${crypto.randomUUID()}.${
+      payload.contentType.split('/')[fileExtensionIndex]
+    }`;
+
+    const putObjectCommand = new PutObjectCommand({
+      Bucket: this.bucketName,
+      Key: fileKey,
+      Body: payload.buffer,
+      ContentType: payload.contentType,
+    } as PutObjectCommandInput);
```
v0ldemar01: Let's create separate service for aws s3

v0ldemar01 (changes requested):

backend/src/packages/files/file.service.ts
Nanaier: @v0ldemar01 quick question - How will we retrieve the file from AWS? Using presigned URL? or is there another way to do that?

backend/src/index.d.ts
```diff
@@ -1,9 +1,11 @@
 import 'fastify';
 
+import { type FileUploadRequestDto } from '#packages/files/files.js';
 import { type UserAuthResponseDto } from '#packages/users/users.js';
 
 declare module 'fastify' {
   interface FastifyRequest {
     user?: UserAuthResponseDto;
+    fileBuff?: FileUploadRequestDto;
```
fshabanov:
```suggestion
    fileBuffer?: FileUploadRequestDto;
```

backend/src/libs/plugins/file-upload/file-upload.plugin.ts
```diff
@@ -0,0 +1,40 @@
...
+  fastify.addHook(ControllerHook.PRE_VALIDATION, async (request) => {
+    if (!request.isMultipart()) {
+      return;
+    }
+
+    const { file } = request.body as { file: MultipartFile };
+
+    if (file.file.truncated) {
+      return;
+    }
```
fshabanov: I suggest we throw a file size error here and set ```throwFileSizeLimit``` to ```false``` in base-server-application.ts
The reason is that with ```throwFileSizeLimit``` set to ```true``` the library will throw an error message, which is not user-friendly.

shared/src/packages/files/libs/types/file-get-all-item-response-dto.type.ts
```diff
@@ -0,0 +1,7 @@
+type FileGetAllItemResponseDto = {
+  id: number;
+  url: string;
+  contentType: string;
```
fshabanov:
```suggestion
  contentType: ValueOf<typeof ContentType>;
```

fshabanov (changes requested):

backend/src/packages/files/file.repository.ts
```diff
@@ -0,0 +1,88 @@
...
+    return FileEntity.initialize({
+      id: file.id,
+      url: file.url,
+      contentType: file.contentType,
+      createdAt: new Date(file.createdAt),
+      updatedAt: new Date(file.updatedAt),
+    });
+  }
+
+  public async findAll(): Promise<FileEntity[]> {
```
OlenaYatseniuk: Do we need this method? I think it's unnecessary to implement it. You can return Promise.resolve() for this.

backend/src/libs/packages/server-application/base-server-application.ts
```diff
@@ -106,13 +110,26 @@ class BaseServerApplication implements ServerApplication {
     );
   }
 
+  private async initMultipart(): Promise<void> {
+    await this.app.register(fastifyMultipart, {
```
liza-veis, line 123: this should be in initPlugins

backend/src/packages/files/aws.service.ts
```diff
@@ -0,0 +1,84 @@
...
+    this.s3Client = new S3Client({
+      region: this.region,
+      credentials: {
+        accessKeyId: this.accessKeyId,
+        secretAccessKey: this.secretAccessKey,
+      },
+    } as S3ClientConfig);
+  }
+
+  public getS3Client(): S3Client {
```
liza-veis:
```suggestion
  public get s3Client(): S3Client {
```

backend/src/packages/files/aws.service.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  public async getPreSignedURL(fileKey: string): Promise<string> {
+    const command = new GetObjectCommand({
+      Bucket: this.bucketName,
+      Key: fileKey,
+    });
+
+    const signedUrl: string = await getSignedUrl(this.s3Client, command, {
+      expiresIn: 3600,
```
liza-veis: create a constant property for this

backend/src/packages/files/aws.service.ts
```diff
@@ -0,0 +1,84 @@
...
+      Bucket: this.bucketName,
+      Key: fileKey,
+      Body: buffer,
+      ContentType: contentType,
+    } as PutObjectCommandInput);
+
+    await this.s3Client.send(putObjectCommand);
+  }
+
+  public async getPreSignedURL(fileKey: string): Promise<string> {
```
liza-veis:
```suggestion
  public async getPreSignedUrl(fileKey: string): Promise<string> {
```

backend/src/packages/files/aws.service.ts
```diff
@@ -0,0 +1,84 @@
...
+    });
+
+    const signedUrl: string = await getSignedUrl(this.s3Client, command, {
+      expiresIn: 3600,
+    });
+
+    return signedUrl;
+  }
+
+  public getURL(fileKey: string): string {
```
liza-veis: we capitalize abbreviations only if it's the first word
```suggestion
  public getUrl(fileKey: string): string {
```

backend/src/packages/files/file.service.ts
```diff
@@ -0,0 +1,92 @@
...
+  public constructor({ fileRepository, awsService }: FileServiceDependencies) {
+    this.fileRepository = fileRepository;
+    this.awsService = awsService;
+  }
+
+  public find(): ReturnType<Service['find']> {
+    return Promise.resolve(null);
+  }
+
+  //FUTURE TODO: CREATE CONTROLLER FOR findById
```
liza-veis: let's start comments without FUTURE, just TODO:, and use usual text after that
```suggestion
  // TODO: create controller for findById
```

backend/src/packages/files/file.service.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  //FUTURE TODO: CREATE CONTROLLER FOR findAll
+
+  public async findAll(): Promise<FileGetAllResponseDto> {
+    const files = await this.fileRepository.findAll();
+
+    return {
+      items: files.map((file) => file.toObject()),
+    };
+  }
```
OlenaYatseniuk: Also, you can return Promise.resolve here.

liza-veis (changes requested):

backend/src/packages/files/files.ts
```diff
@@ -0,0 +1,34 @@
...
+export { type AWSService } from './aws.service.js';
+export { FileModel } from './file.model.js';
+export { type FileService } from './file.service.js';
+export {
+  type AWSUploadRequestDto,
+  type FileCommonQueryResponse,
+  type FileCreateQueryPayload,
+  type FileUploadRequestDto,
+  type FileUploadResponseDto,
+} from './libs/types/types.js';
```
liza-veis: we don't need to export everything, just things that we use

liza-veis (changes requested):

shared/src/packages/files/libs/types/file-get-all-response-dto.type.ts
```diff
@@ -0,0 +1,7 @@
+import { type FileGetAllItemResponseDto } from './types.js';
+
+type FileGetAllResponseDto = {
+  items: FileGetAllItemResponseDto[];
+};
+
+export { type FileGetAllResponseDto };
```
OlenaYatseniuk: So, this type you also don't need.

OlenaYatseniuk (changes requested):
You did a great job!

backend/src/packages/files/file.repository.ts
Nanaier: since files is a separate table - we might need it in the future. I won't delete It for now, but I agree that the code is redundant a bit...

OlenaYatseniuk: So, I think we need coaches' thoughts about it, because I implemented the surveys endpoint and got the same comment from @Liza-Veis here https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/66#discussion_r1306434148

Liza-Veis: yep, we don't need to implement methods we don't use

Liza-Veis: but let's leave findById, as we will use it in next step

liza-veis: let's remove get all method

OlenaYatseniuk: Yes, is it about both findById and findAll?
Because, it seems to me, that findAll won't be used in future.  
If I am mistaken, please, correct me.

liza-veis: Yes, only findAll

liza-veis:
Good job 🙌  just remove findAll from repository level too

backend/src/packages/files/aws.service.ts
```diff
@@ -0,0 +1,98 @@
...
+    await this.s3Client.send(putObjectCommand);
+  }
+
+  public async getPreSignedUrl(fileKey: string): Promise<string> {
+    const command = new GetObjectCommand({
+      Bucket: this.bucketName,
+      Key: fileKey,
+    });
+
+    const minuteInSec = 3600;
```
v0ldemar01: Let's put into constants

backend/src/packages/files/aws.service.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    const signedUrl: string = await getSignedUrl(this.s3Client, command, {
+      expiresIn: minuteInSec,
+    });
+
+    return signedUrl;
+  }
+
+  public getUrl(fileKey: string): string {
+    return `https://${this.bucketName}.s3.${this.region}.amazonaws.com/${fileKey}`;
```
v0ldemar01: Create a helper to generate from template

backend/src/packages/files/aws.service.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  public getFileKey(url: string): string | undefined {
```
v0ldemar01: Where is it used?

v0ldemar01 (changes requested):
Nice job, please fix small comments)

backend/src/packages/files/aws.service.ts
Nanaier: the idea is to use it in the future when a person needs preSigned URL - they pass URL here to get fileKey and use It later on in the preSigned URL generation. I can delete if needed tho...

v0ldemar01: Yes, please delete for now) The regexp is fine but the logic can be replaced with `location.split('s3.amazonaws.com/').pop()`

backend/src/packages/files/aws.service.ts
```diff
@@ -0,0 +1,98 @@
...
+import { type AWSUploadRequestDto } from './libs/types/types.js';
+
+type AWSServiceDependencies = {
+  region: string;
+  accessKeyId: string;
+  secretAccessKey: string;
+  bucketName: string;
+};
+
+class AWSService {
```
v0ldemar01: AWS is not consist of only S3)

backend/src/packages/files/aws.service.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  public constructor({
+    region,
+    accessKeyId,
+    secretAccessKey,
+    bucketName,
+  }: AWSServiceDependencies) {
+    this.region = region;
+    this.accessKeyId = accessKeyId;
+    this.secretAccessKey = secretAccessKey;
```
v0ldemar01: We don't need to store it

backend/src/packages/files/aws.service.ts
```diff
@@ -0,0 +1,98 @@
...
+    const match = url.match(regex);
+    const fileKeyIndex = 1;
+
+    if (!match) {
+      return;
+    }
+
+    return match[fileKeyIndex];
+  }
+}
```
v0ldemar01: Please also add error handling like 
```js
private throwError(err: unknown): never {
    if (err instanceof S3ServiceException) {
      throw new FilesError({
        message: err.message as ExceptionMessage,
        status: err.$response?.statusCode,
      });
    }
    throw new FilesError();
  }
```

v0ldemar01 (changes requested):

OlenaYatseniuk (dismissed):

backend/src/packages/files/aws.service.ts
Nanaier: should I rename the service to S3 then?

backend/src/packages/files/aws.service.ts
Nanaier: I added an AWSError - I hope that was what you wanted

backend/src/packages/files/libs/helpers/get-url.ts
```diff
@@ -0,0 +1,18 @@
+function generateUrl(
+  template: string,
+  placeholders: Record<string, string>,
+): string {
+  let url = template;
+  for (const key in placeholders) {
+    if (Object.hasOwnProperty.call(placeholders, key)) {
```
HIMDanny: Maybe let's use `Object.hasOwn`, as it is a recommended over `hasOwnProperty()`? Or even write it with `Object.keys()` instead of for...in
```suggestion
    if (Object.hasOwn(placeholders, key)) {
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

Nanaier: hey, thanks, for suggestion - updated my code!

HIMDanny: now you can even remove the `Object.hasOwn()` condition, as `Object.keys()` iterates only over properties that are attached to the object itself

Nanaier: yep, you are right - made changes in a hurry ;)

HIMDanny (dismissed):

backend/src/packages/files/aws.service.ts
```diff
@@ -0,0 +1,86 @@
+import {
```
dmytro-kucherenko: I think it's better to store aws service (or not aws as @v0ldemar01 said) in libs/packages/aws folder, like jwt

dmytro-kucherenko (commented):
Seems pretty good

backend/src/packages/files/aws.service.ts
Nanaier: @v0ldemar01 could you clarify this requested change + look into other changes

iamAlinaaa (dismissed):
looks good with all changes

backend/src/packages/files/aws.service.ts
```diff
@@ -0,0 +1,86 @@
...
+  GetObjectCommand,
+  PutObjectCommand,
+  type PutObjectCommandInput,
+  S3Client,
+  type S3ClientConfig,
+} from '@aws-sdk/client-s3';
+import { getSignedUrl } from '@aws-sdk/s3-request-presigner';
+
+import { MINUTE_IN_SEC } from './libs/constants/constants.js';
+import { generateUrl } from './libs/helpers/get-url.js';
```
fshabanov:
```suggestion
import { generateUrl } from './libs/helpers/helpers.js';
```

backend/src/packages/files/libs/constants/constants.ts
```diff
@@ -0,0 +1,3 @@
+const MINUTE_IN_SEC = 3600;
```
fshabanov: Did you mean seconds in an hour?

backend/src/packages/files/libs/helpers/get-url.ts
```diff
@@ -0,0 +1,17 @@
+function generateUrl(
```
fshabanov: Use arrow function. Also, the name of the file and the name of the function should be the same

shared/src/libs/exceptions/aws-error/aws-error.exception.ts
```diff
@@ -0,0 +1,22 @@
+import { HTTPCode } from '#libs/packages/http/http.js';
+import { type ValueOf } from '#libs/types/value-of.type.js';
```
fshabanov:
```suggestion
import { type ValueOf } from '#libs/types/types.js';
```

fshabanov (changes requested):

backend/src/packages/files/aws.service.ts
v0ldemar01: @Nanaier So, we need to catch aws oriented error that is linked to aws action and throw our custom error. S3ServiceException is AWS linked error.

backend/src/packages/files/libs/helpers/get-url.ts
v0ldemar01, line 1: The name should contain .helper suffix

backend/src/packages/files/libs/helpers/get-url.ts
```diff
@@ -0,0 +1,17 @@
+const getUrl = (
```
v0ldemar01: Let's create a shared helper for such purpose
```ts
const replaceTemplateWithValue = (template: string, replacements: Record<string, string | number>): string => {
	return template.replace(/\{\{([^}]+)\}\}/g, (_, placeholder) => String(replacements[placeholder]) ?? '');
};
```

v0ldemar01 (changes requested):

shared/src/libs/enums/exception-message.enum.ts
```diff
@@ -4,6 +4,8 @@ const ExceptionMessage = {
   USER_NOT_FOUND: 'User with these credentials was not found.',
   INCORRECT_CREDENTIALS: 'Incorrect credentials.',
   USER_ALREADY_EXISTS: 'User already exists.',
+  INCORRECT_FILE_TYPE: 'Content type of the file is not in png, jpeg.',
```
OlesiaTr: I think it could be improved a bit for better clarity:
`INCORRECT_FILE_TYPE: 'Content type of the file is not in PNG or JPEG',` or
`INCORRECT_FILE_TYPE: 'Content type of the file is not in PNG/JPEG',`

dmytro-kucherenko (dismissed):

backend/src/libs/packages/server-application/base-server-application.ts
```diff
@@ -1,19 +1,23 @@
...
+import { APIPath, ContentType, ServerErrorType } from '#libs/enums/enums.js';
 import { type ValidationError } from '#libs/exceptions/exceptions.js';
 import { type Config } from '#libs/packages/config/config.js';
 import { type Database } from '#libs/packages/database/database.js';
 import { jwtService } from '#libs/packages/jwt/jwt.js';
 import { type Logger } from '#libs/packages/logger/logger.js';
-import { authorization as authorizationPlugin } from '#libs/plugins/plugins.js';
+import {
+  authorization as authorizationPlugin,
+  fileUpload as fileUploadingPlugin,
```
fshabanov:
```suggestion
  fileUpload as fileUploadPlugin,
```

backend/src/packages/files/aws.service.ts
fshabanov: I think you can name it file service

backend/src/packages/files/aws.service.ts
```diff
@@ -0,0 +1,105 @@
...
+    return signedUrl;
+  }
+
+  public getUrl(fileKey: string): string {
+    return replaceTemplateWithValue(
+      'https://{bucket}.s3.{region}.amazonaws.com/{fileKey}',
+      {
+        bucket: this.bucketName,
+        region: this.region,
+        fileKey: fileKey,
```
fshabanov:
```suggestion
        fileKey,
```

shared/src/libs/exceptions/file-error/file-error.exception.ts
```diff
@@ -0,0 +1,22 @@
+import { HTTPCode } from '#libs/packages/http/http.js';
+import { type ValueOf } from '#libs/types/value-of.type.js';
```
fshabanov:
```suggestion
import { type ValueOf } from '#libs/types/types.js';
```

shared/src/packages/files/libs/types/aws-upload-request-dto.ts
```diff
@@ -0,0 +1,9 @@
+import { type ContentType, type ValueOf } from '#index.js';
+
+type AWSUploadRequestDto = {
```
fshabanov:
```suggestion
type AWSFileUploadRequestDto = {
```

fshabanov (changes requested):

backend/src/packages/files/aws.service.ts
Nanaier: @fshabanov but I already have file.service, is there another name I should go with?

liza-veis: let's name it `S3` and move to libs/packages/s3/s3.package.ts and then re-export from libs/packages/file/file.ts

backend/src/libs/packages/s3/s3.package.ts
```diff
@@ -16,14 +16,14 @@ import { type ValueOf } from '#libs/types/types.js';
...
-type AWSServiceDependencies = {
+type S3PackageDependencies = {
   region: string;
   accessKeyId: string;
   secretAccessKey: string;
   bucketName: string;
 };
 
-class AWSService {
+class S3Package {
```
liza-veis: just s3 without package
```suggestion
class S3 {
```

backend/src/libs/packages/s3/libs/types/types.ts
```diff
@@ -0,0 +1 @@
+export { type AWSFileUploadRequestDto } from 'shared/build/index.js';
```
liza-veis:
```suggetion
export { type FileUploadRequestDto } from 'shared/build/index.js';
```

liza-veis (changes requested):

backend/src/libs/packages/s3/libs/types/types.ts
Nanaier: we have the exact same type name already - so I'll use S3FileUploadRequestDto

liza-veis (dismissed):

shared/src/packages/files/libs/types/file-get-all-item-response-dto.type.ts
```diff
@@ -0,0 +1,9 @@
+import { type ContentType, type ValueOf } from '#index.js';
```
fshabanov: I think this will cause circular dependency

shared/src/packages/files/libs/types/file-upload-request-dto.type.ts
```diff
@@ -0,0 +1,8 @@
+import { type ContentType, type ValueOf } from '#index.js';
```
fshabanov: Same here

shared/src/packages/files/libs/types/s3-file-upload-request-dto.ts
```diff
@@ -0,0 +1,9 @@
+import { type ContentType, type ValueOf } from '#index.js';
```
fshabanov: Same

fshabanov (changes requested):

shared/src/packages/files/libs/types/s3-file-upload-request-dto.ts
Nanaier: @fshabanov added the requested changes!

liza-veis (approved):

BrianSammit (approved):

shared/src/libs/helpers/replace-template-with-value/replace-template-with-value.helper.ts
```diff
@@ -0,0 +1,13 @@
+const replaceTemplateWithValue = (
+  template: string,
+  replacements: Record<string, string | number>,
+): string => {
+  let result = template;
+  for (const [placeholder, value] of Object.entries(replacements)) {
+    result = result.replaceAll(`{${placeholder}}`, String(value));
+  }
```
v0ldemar01, line 8: Let's use .reduce here

v0ldemar01 (changes requested):
LGTM, but fix last comment please)

shared/src/libs/helpers/replace-template-with-value/replace-template-with-value.helper.ts
Nanaier: @v0ldemar01 - we have a rule against using it `Array#reduce()` is not allowed eslint[unicorn/no-array-reduce]
Screenshot 2023-09-01 at 10 16 52

v0ldemar01: Got it, sorry

v0ldemar01 (approved):

PR #95, opened 2023-08-28
cp-81: Profile settings layout

HIMDanny:
- Created a common switcher (checkbox) component
- Created setting component
- Created profile settings page layout

To test, please add a profile settings component in the router provider

## Images

Mobile view:

image

Desktop view:

image

Toggled-off switcher:

image


frontend/src/pages/profile-settings/profile-settings.tsx
```diff
@@ -0,0 +1,28 @@
+import { Switch } from '#libs/components/components.js';
+import { useAppForm } from '#libs/hooks/hooks.js';
+
+import { Setting } from './components/components.js';
+import styles from './styles.module.scss';
+
+const ProfileSettings: React.FC = () => {
+  const { control } = useAppForm({ defaultValues: { notification: true } });
```
danilittaa: I think you should move default values in libs/constants.ts as it's done for sign up for example

HIMDanny: I was thinking about this, but not sure. Let's also ask @Liza-Veis, @v0ldemar01, @fshabanov

iamAlinaaa (dismissed):
all is good, waiting for descision about default values in libs/constants

frontend/src/pages/profile-settings/profile-settings.tsx
fshabanov: I agree with @danilittaa, move default values to a separate file

HIMDanny: no problem

frontend/src/libs/components/switch/switch.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const Switch = <T extends FieldValues>({
+  control,
+  name,
+  label,
+}: Properties<T>): JSX.Element => {
+  const { field } = useFormController({ name, control });
+
+  return (
+    <label className={styles['container']} htmlFor={name}>
```
liza-veis: we don't need htmlFor and id property for input when we nest in label

liza-veis (changes requested):

Nanaier (dismissed):

Nanaier:
hey, quick question - should you link your issue to your PR though? I feel like we shouldn't be doing that ...

frontend/src/libs/components/switch/switch.tsx
```diff
@@ -0,0 +1,37 @@
+import {
+  type Control,
+  type FieldPath,
+  type FieldValues,
+} from 'react-hook-form';
```
v0ldemar01: Let's move to common types,  but add a prefix Form

v0ldemar01: The main idea is that if we decide to remove that lib, so it would be easier to handle it in one place

v0ldemar01 (changes requested):

HIMDanny:
> hey, quick question - should you link your issue to your PR though? I feel like we shouldn't be doing that ...

yep, you're right)

frontend/src/libs/components/switch/switch.tsx
HIMDanny: ok, no problem. In the input component types are imported the same way, should I change them as well?

frontend/src/libs/components/switch/switch.tsx
HIMDanny: right) Will be changed

frontend/src/libs/components/switch/switch.tsx
fshabanov: Yes, change all imports for these types to be from common types

frontend/src/libs/hooks/hooks.ts
```diff
@@ -2,5 +2,11 @@ export { useAppDispatch } from './use-app-dispatch/use-app-dispatch.hook.js';
 export { useAppForm } from './use-app-form/use-app-form.hook.js';
 export { useAppSelector } from './use-app-selector/use-app-selector.hook.js';
 export { useCallback, useEffect, useMemo, useState } from 'react';
-export { useController as useFormController } from 'react-hook-form';
+export {
+  type Control as FormControl,
+  type FieldErrors as FormFieldErrors,
+  type FieldPath as FormFieldPath,
+  type FieldValues as FormFieldValues,
```
fshabanov: But these are not hooks
Move these imports to libs/types folder

fshabanov (changes requested):

frontend/src/libs/hooks/hooks.ts
HIMDanny: oh, right... My bad

frontend/src/libs/types/types.ts
```diff
@@ -1,4 +1,10 @@
 export { type AsyncThunkConfig } from './async-thunk-config.type.js';
+export {
+  type Control as FormControl,
+  type FieldErrors as FormFieldErrors,
+  type FieldPath as FormFieldPath,
+  type FieldValues as FormFieldValues,
+} from 'react-hook-form';
```
dmytro-kucherenko: Do we really need these reexports?

frontend/src/libs/components/switch/switch.tsx
```diff
@@ -0,0 +1,35 @@
...
+}: Properties<T>): JSX.Element => {
+  const { field } = useFormController({ name, control });
+
+  return (
+    <label className={styles['container']}>
+      <span className="visually-hidden">{label}</span>
+      <input
+        className={styles['input']}
+        type="checkbox"
+        checked={field.value}
```
dmytro-kucherenko, line 29: I think form control will work without `checked={field.value}`, but I could be wrong

frontend/src/libs/types/types.ts
HIMDanny: it was not my idea) Please, check https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/95#discussion_r1308234923

dmytro-kucherenko: Okay, I see

frontend/src/libs/components/switch/switch.tsx
HIMDanny: checkbox input needs checked property to be controlled. It does not understand the value prop. Also without this prop we cannot set the default value from `useAppForm` in the checkbox input

frontend/src/libs/types/types.ts
dmytro-kucherenko: It's also suggested to do it in separate folder

image

frontend/src/libs/components/switch/switch.tsx
dmytro-kucherenko: Okay, you good

liza-veis (approved):

fshabanov (approved):

v0ldemar01 (approved):

PR #104, opened 2023-08-29
cp-98: Add hover effect on elements in sign up page

HIMDanny:
- Hover effect on the submit button
- Added a new shade of gray so the hover on the button looks nicer
- Hover effect on the sign-in link on the signup page

## Images

Submit button hover:

image
image

Sign in link hover:

image
image

### DIFF
--- a/frontend/src/assets/css/variables.scss
+++ b/frontend/src/assets/css/variables.scss
@@ -4,6 +4,7 @@
   --black: #000000;
   --gray-100: #e2e3e5;
   --gray-200: #6c707a;
+  --gray-250: #484b53;
   --gray-300: #302e36;
   --blue-100: #ebf3ff;
   --blue-200: #acb4d0;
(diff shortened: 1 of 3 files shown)


frontend/src/assets/css/variables.scss
```diff
@@ -4,6 +4,7 @@
   --black: #000000;
   --gray-100: #e2e3e5;
   --gray-200: #6c707a;
+  --gray-250: #484b53;
```
liza-veis: let's either use gray-300, all we need to switch them to have gray-300, gray-400

liza-veis (changes requested):

frontend/src/assets/css/variables.scss
HIMDanny: did I understand you correctly: change current gray-300 to gray-400 and replace in all styles to it, and change gray-250 to gray-300?

liza-veis: yep

fshabanov (approved):

liza-veis (approved):

PR #110, opened 2023-08-29
cp-91:  Add sidebar css to follow the scroll

Ekit18:
image


frontend/src/libs/components/sidebar/styles.module.scss
```diff
@@ -18,10 +18,13 @@
 }
 
 .sidebar {
+  position: sticky;
+  top: 0;
   display: flex;
   flex-direction: column;
   gap: 262px;
   width: 88px;
+  height: 100vh;
```
liza-veis: We need to avoid vh, it may cause issues on mobile, try to use 100%

liza-veis (changes requested):

Xxandder (dismissed):

frontend/src/libs/components/sidebar/styles.module.scss
Ekit18: @Liza-Veis it doesn`t work with 100%.
image

frontend/src/libs/components/sidebar/styles.module.scss
```diff
@@ -18,10 +18,13 @@
 }
 
 .sidebar {
+  position: sticky;
+  top: 0;
   display: flex;
   flex-direction: column;
   gap: 262px;
```
dmytro-kucherenko, line 24: @Liza-Veis Maybe it's also an option to make appropriate icons centering in this issue?

image

OlesiaTr (dismissed):

frontend/src/libs/components/sidebar/styles.module.scss
HIMDanny: for me it works without height property at all)

image

Ekit18: @HIMDanny put smth big to have scroll

frontend/src/libs/components/sidebar/styles.module.scss
liza-veis: what do you mean by icons centering? if you mean that they should be at the center of the screen, from design it looks like they are not centered vertically

frontend/src/libs/components/sidebar/styles.module.scss
HIMDanny: now I see. Can we use `position: fixed` and `height: 100%`? It seems to be working

Ekit18: if i change to position fixed i have to make margin left 
image

HIMDanny: you are right. But it is one of the working solutions. I've seen this on a lot of websites so I think this is not so bad

Ekit18: @Liza-Veis what is the right way?

liza-veis: will overflow-y hidden fix this (with position: sticky, height  100%) ? using vh for sidebar will create another issue later, that's why we can't use it

Ekit18: @Liza-Veis height 100% doesn`t work. only vh works with position sticky. height 100% works with position fixed

OlenaYatseniuk (dismissed):

frontend/src/libs/components/sidebar/styles.module.scss
dmytro-kucherenko: Okay, I realized it. It may cause lack of responsiveness, but if it's needed this way, no more comments)

frontend/src/libs/components/sidebar/styles.module.scss
liza-veis: ok, got it, then let's use position fixed with padding. But pay attention padding should be in sidebar component, not in the app or somewhere else

frontend/src/libs/components/sidebar/styles.module.scss
liza-veis: we will think about it a bit later, let's leave it for now

dmytro-kucherenko (approved):

fshabanov (approved):

BrianSammit (approved):

Vika030718 (approved):

iamAlinaaa (approved):

liza-veis (approved):

PR #117, opened 2023-08-30
cp-112: Sign in response

HIMDanny:
- Changed UserSignInResponseDto and UserAuthResponseDto
- Sign-up response returns user object without password data
- Same response from sign-in and sign-up endpoints

### Sign-up

image

### Sign-in

image


shared/src/packages/users/libs/types/user-sign-in-response-dto.type.ts
```diff
@@ -1,8 +1,8 @@
+import { type UserAuthResponseDto } from './types.js';
```
v0ldemar01: Please avoid circular dependency

v0ldemar01 (changes requested):
Is swagger config suitable for new response?

shared/src/packages/users/libs/types/user-sign-in-response-dto.type.ts
HIMDanny: I was looking at the `user-sign-up-response-dto.ts` file. Should I change the import there as well?

OlesiaTr (dismissed):

shared/src/packages/users/libs/types/user-sign-in-response-dto.type.ts
fshabanov: @HIMDanny yes, please, update it as well

v0ldemar01: Sure, as a solution: create value-of.type.ts file as in shared, but export { ValueOf } from 'shared/build/index.js' here

HIMDanny: was it meant for me?) Because I'm not using `ValueOf`)

backend/src/packages/auth/auth.controller.ts
```diff
@@ -143,7 +143,7 @@ class AuthController extends BaseController {
 
     return {
```
liza-veis, line 171: update swagger docs comment above please as @v0ldemar01 said

HIMDanny:
> Is swagger config suitable for new response?

good point! Will configure it

liza-veis (changes requested):

liza-veis (approved):

v0ldemar01 (approved):

PR #119, opened 2023-08-30
cp-115: Add try catch to authorization token decode

Ekit18:
image
image


OlesiaTr (dismissed):

backend/src/libs/plugins/authorization/authorization.plugin.ts
```diff
@@ -38,16 +38,21 @@ const authorization = fp<Options>((fastify, { services }, done) => {
...
+      if (!authorizedUser) {
+        throw new AuthError({
+          message: ExceptionMessage.INVALID_TOKEN,
+        });
+      }
+
+      request.user = authorizedUser;
+    } catch {
       throw new AuthError({
         message: ExceptionMessage.INVALID_TOKEN,
```
v0ldemar01: Let's change message to ExceptionMessage.UNAUTHORIZED_USER

v0ldemar01 (changes requested):

backend/src/libs/plugins/authorization/authorization.plugin.ts
Ekit18: done

liza-veis (approved):

PR #139, opened 2023-08-31
cp-19: Fix error is shown when unauthenticated user opens the web app for the first time

OlenaYatseniuk:
- Added handleGetAuthenticatedUser function which check if token exists and dispatch getAuthenticatedUser action
image


v0ldemar01 (dismissed):

frontend/src/libs/components/app/app.tsx
```diff
@@ -13,10 +15,18 @@ const App: React.FC = () => {
     authenticatedUserDataStatus: auth.authenticatedUserDataStatus,
   }));
 
-  useEffect(() => {
-    void dispatch(authActions.getAuthenticatedUser());
+  const handleGetAuthenticatedUser = useCallback(async (): Promise<void> => {
+    const hasToken = await storage.has(StorageKey.TOKEN);
```
liza-veis: let's do this in getAuthenticatedUser action directly and return null if there is no token, we can't use services inside components

liza-veis (changes requested):

liza-veis (approved):

PR #141, opened 2023-08-31
cp-128: [BUG] Error is displayed on the 'Sign Up' screen after redirection from 'Sign in' screen#128

BrianSammit:
# Changes 

- Check the presence of a token in user action and auth action to fix the toast message  'User is not is authorized' that is displayed when the app first load and move to the sign-up page.

Screenshot 2023-08-31 183255
Screenshot 2023-08-31 183316


liza-veis (approved):

iamAlinaaa (approved):

Nanaier (approved):

iamAlinaaa (commented):
hope to test it as soon as possible

fshabanov (approved):

Issue #54, opened 2023-08-22
Chat screen


liza-veis:
- [ ] Implement chat screen
- [ ] Use mocked data for chat items (leave at least one for testing)
- [ ] Add search by name (should be case insensitive)

![Chat](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/64521658/fc104e96-c6a5-4400-b3cf-074ec53b175e)

User story:
As a user, I want to see chat items and be able to use search on the Chat screen so that I can quickly find the needed conversatins by name.

Acceptance criteria:
1. Chat items should be displayed on the Chat screen
2. User should be able to search specific chat by entering input to the search field on the Chat screen
3. Search must be case insensitive

Vika030718:
I have created a chat screen, now it looks like this.

Note 1:  We have changed the word 'topic' to the word 'chat' in the search field and in the link. This was approved by coaches. 
Note 2: Arrows were removed from the chat item on purpose.

![Image](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/4521937/c4c8d0fd-f441-4fe3-a1d9-ba14dbbd1869)

To QA Meam: It also has a bug for now - the title should be Chat, but not Chat List. Could you please create a ticket for this?

PR #114, opened 2023-08-30
cp-78: Add meditation player layout

OlenaYatseniuk:
Changes:

- Add meditation player layout according to Figma and requirements (mocked data for now)
- User is able to play audio by URL (all functionality works)
image

PR #154, opened 2023-09-02
cp-113: Add Mobile Profile Settings

iamAlinaaa:
1. Created Common checkbox component and Setting component
2. Added profile button to header #54 
3. Improvements to reusable header
4. When user presses profile button, they should be redirected to profile screen with only one item for now (Notifications and Reminders)
5.  After user presses "Notifications and Reminders" button, they should be redirected to settings page (leave only Allow Notifications setting for now)
6. Added navigation to previous screen

Снимок экрана 2023-09-02 191653
Снимок экрана 2023-09-04 123205
Снимок экрана 2023-09-04 123213

https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/111873795/273cb802-e87b-48e1-845b-87525d0af649


mobile/src/screens/chat-list/chat-list.tsx
```diff
@@ -1,41 +1,48 @@
...
+  Header,
   InputSearch,
   Link,
   ScrollView,
-  Text,
   View,
 } from '#libs/components/components';
 import { AppColor, RootScreenName } from '#libs/enums/enums';
-import { useEffect, useNavigation, useSearch } from '#libs/hooks/hooks';
+import { ChatScreenName } from '#libs/enums/navigation/chat-screen-name.enum';
```
danilittaa:
```suggestion
import { ChatScreenName } from '#libs/enums/enums';
```

danilittaa (changes requested):

mobile/src/libs/components/icon-button/icon-button.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+import { TouchableOpacity } from '#libs/components/components';
+
+import { styles } from './styles';
+
+type IconButtonProperties = {
```
Vika030718:
```suggestion
type Properties = {
```

mobile/src/screens/profile/components/profile-category/profile-category.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+import { Pressable, Text, View } from '#libs/components/components';
+
+import { styles } from './styles';
+
+type ProfileCategoryProperties = {
```
Vika030718:
```suggestion
type Properties = {
```

mobile/src/libs/components/switch/switch.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+
+import { RNSwitch, View } from '#libs/components/components';
```
fshabanov: We can import RNSwitch directly from react-native for this file, no need to re-export it. Because that component will be used only in this file

mobile/src/libs/components/switch/switch.tsx
```diff
@@ -0,0 +1,43 @@
...
+          false: styles.switchFalse.backgroundColor,
+          true: styles.switchTrue.backgroundColor,
+        }}
+        thumbColor={
+          field.value ? styles.switchTrue.color : styles.switchFalse.color
+        }
+        ios_backgroundColor={styles.switchTrue.backgroundColor}
+        onValueChange={(value): void => {
+          field.onChange(value);
+        }}
```
fshabanov:
```suggestion
        onValueChange={onChange}
```
And destructure field variable:
`const { onChange, value } = field;`

mobile/src/screens/chat-list/chat-list.tsx
```diff
@@ -54,7 +61,13 @@ const ChatList: React.FC = () => {
         />
         <ScrollView contentContainerStyle={styles.list}>
           {filteredChats.map((item) => (
-            <Card title={item.title} onPress={onPress} key={item.id} />
+            <Card
+              title={item.title}
+              onPress={(): void => {
+                onPress(item.title);
+              }}
```
fshabanov: Avoid inline functions

mobile/src/screens/profile-settings/libs/constants.ts
```diff
@@ -0,0 +1,5 @@
+const DEFAULT_NOTIFICATION_SETTINGS_VALUES = {
```
fshabanov:
```suggestion
const NOTIFICATION_SETTINGS_DEFAULT_VALUE = {
```

mobile/src/screens/profile-settings/profile-settings.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import LinearGradient from 'react-native-linear-gradient';
```
fshabanov: I see that we are using this component in lots of files, let's create a custom component for that and use it everywhere

mobile/src/screens/profile-settings/profile-settings.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const { control } = useAppForm({
+    defaultValues: DEFAULT_NOTIFICATION_SETTINGS_VALUES,
+  });
+
+  return (
+    <LinearGradient
+      colors={[AppColor.WHITE, AppColor.BLUE_100]}
+      style={styles.linearGradient}
+    >
+      <Header navigation={navigation} title={route.name} isArrowVisible />
```
fshabanov: In Header component, let's use navigation from hook (`const navigation = useNavigation()` and make title property optional, if it is not passed, use `const { name } = useAppRoute();`

mobile/src/screens/profile/profile.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { type NavigationScreenProperties } from '#libs/types/types';
+
+import { ProfileCategory } from './components/components';
+import { styles } from './styles';
+
+const Profile = ({
+  navigation,
+  route,
+}: NavigationScreenProperties): JSX.Element => {
+  const onPress = (): void => {
```
fshabanov:
```suggestion
  const handleCategoryPress = (): void => {
```

fshabanov (changes requested):

liza-veis:
Make items wider and more similar to chat items. Use Card component for notification items because it has the same layout and make setting switch the same size as Card component

OlenaYatseniuk (dismissed):
You did a great job!

mobile/src/assets/img/icons/user.svg
danilittaa, line 1: I think you should change color to 'currentColor' for all SVGs so that it can be easily changed later

mobile/src/libs/components/icon-button/icon-button.tsx
iamAlinaaa: i dont think we should strictly name them only Properties 🤔

danilittaa: @iamAlinaaa, but we consistently name the props 'Properties' everywhere. I think you should rename it

mobile/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,136 @@
...
+//       {isSettingsVisible && (
+//         <View style={styles.settingsContainer}>
+//           <IconButton onPress={handleIconPress} iconSourceSvg={<UserIcon />} />
+//         </View>
+//       )}
+//     </View>
+//   );
+// };
+
+// export { Header };
```
dmytro-kucherenko: Oh, what is these comments doing here?)

iamAlinaaa: @Dmytro-Kucherenko im still in progress of fixing comments and kept the old logic in case i will need somenthing from there

mobile/src/libs/components/header/header.tsx
```diff
@@ -0,0 +1,72 @@
+import { type NavigationProp } from '@react-navigation/native';
+import React from 'react';
+
+import BackArrowImage from '#assets/img/icons/back-arrow.svg';
+import UserIcon from '#assets/img/icons/user.svg';
```
fshabanov: Use Icon component

mobile/src/libs/components/icon-button/icon-button.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+
+import { TouchableOpacity } from '#libs/components/components';
+
+import { styles } from './styles';
+
+type Properties = {
+  onPress: () => void;
+  iconSourceSvg: JSX.Element;
+};
```
fshabanov: Use Icon component here as well. Instead of iconSourceSvg pass name property

mobile/src/libs/enums/navigation/main-screen-name.enum.ts
```diff
@@ -1,6 +1,6 @@
 const MainScreenName = {
   HOME: 'Home',
-  CHAT_LIST: 'Chat list',
+  CHAT_LIST: 'Chat',
```
fshabanov:
```suggestion
  CHAT: 'Chat',
```

mobile/src/screens/profile-settings/profile-settings.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { useAppForm } from '#libs/hooks/hooks';
+import { type NavigationScreenProperties } from '#libs/types/types';
+
+import { Setting } from './components/components';
+import { NOTIFICATION_SETTINGS_DEFAULT_VALUE } from './libs/constants';
+import { styles } from './styles';
+
+const ProfileSettings = ({
+  route,
+}: NavigationScreenProperties): JSX.Element => {
```
fshabanov: Uer React.FC
Also, we don't need `route` here. Header component uses route name by default

mobile/src/screens/profile/profile.tsx
```diff
@@ -0,0 +1,37 @@
...
+  navigation,
+  route,
+}: NavigationScreenProperties): JSX.Element => {
+  const handleCategoryPress = (): void => {
+    navigation.navigate(RootScreenName.SETTINGS);
+  };
+
+  return (
+    <LinearGradient>
+      <Header title={route.name} isArrowVisible />
```
fshabanov: No need to pass title

mobile/src/screens/profile/profile.tsx
```diff
@@ -0,0 +1,37 @@
...
+} from '#libs/components/components';
+import { AppColor, RootScreenName } from '#libs/enums/enums';
+import { type NavigationScreenProperties } from '#libs/types/types';
+
+import { styles } from './styles';
+
+const Profile = ({
+  navigation,
+  route,
+}: NavigationScreenProperties): JSX.Element => {
```
fshabanov: Use React.FC

fshabanov (changes requested):

mobile/src/libs/components/header/header.tsx
iamAlinaaa: @fshabanov could you please explain about Icon component. I dont have one for now, should i create it and add all SVGs there with purpose to use name property (create enum of existing SVGs and call required using Template literal) as you wrote here "Instead of iconSourceSvg pass name property". Or what should be inside icon component?

OlenaYatseniuk: @iamAlinaaa I quess, you can [see]( https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/114) in my pr what does @fshabanov mean about Icon component.

iamAlinaaa: @OlenaYatseniuk it seems pretty nice, i`ll create the same , thank you 🤩

OlenaYatseniuk: @iamAlinaaa You are welcome=)

fshabanov: @OlenaYatseniuk thanks for sharing!
@iamAlinaaa that is exactly what I meant

ossavemu:
Good Job!

ossavemu (dismissed):

BrianSammit (dismissed):

iamAlinaaa:
I ve written these working changes this way, because i had:
- some linter errors but the app was working as needed
- when i solved linter errors my app had internal error (like something wrong with imports [Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. Imports], but in VS code all was good).
- when i didn have errors neither in linter nor in VS code my app wasnt working as expected (icons just didnt appear)
so after couple of hours or solving this problem i did this and its finaly fulfilled all conditions

OlesiaTr (dismissed):

mobile/src/libs/components/header/header.tsx
```diff
@@ -1,32 +1,73 @@
...
+      )}
+      <View style={styles.titleBadgeContainer}>
+        <Text style={styles.title}>{title ?? name}</Text>
+        {hasValue && <Badge count={badgeCount} />}
+      </View>
+      {isSettingsVisible && (
+        <View style={styles.settingsContainer}>
+          <IconButton
+            onPress={handleIconPress}
+            iconSourceSvg={<Icon name={'user'} color={AppColor.BLUE_300} />}
```
fshabanov:
```suggestion
            iconSourceSvg={<Icon name="user" color={AppColor.BLUE_300} />}
```

mobile/src/navigations/main/main.tsx
```diff
@@ -27,16 +25,23 @@ const Main: React.FC = () => {
         name={MainScreenName.HOME}
         component={Home}
         options={{
-          tabBarIcon: HomeIcon,
+          tabBarIcon: ({ color }): JSX.Element => {
+            return <Icon name={'home'} color={color} />;
```
fshabanov:
```suggestion
            return <Icon name='home' color={color} />;
```

mobile/src/navigations/main/main.tsx
```diff
@@ -27,16 +25,23 @@ const Main: React.FC = () => {
...
         }}
       />
       <BottomTab.Screen
-        name={MainScreenName.CHAT_LIST}
+        name={MainScreenName.CHAT}
         component={Chat}
-        options={{ tabBarIcon: ChatIcon, headerShown: false }}
+        options={{
+          tabBarIcon: ({ color }): JSX.Element => {
+            return <Icon name={'chat'} color={color} />;
```
fshabanov:
```suggestion
            return <Icon name='chat' color={color} />;
```

mobile/src/screens/profile-settings/profile-settings.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const { control } = useAppForm({
+    defaultValues: NOTIFICATION_SETTINGS_DEFAULT_VALUE,
+  });
+
+  return (
+    <LinearGradient>
+      <Header isArrowVisible />
+      <ScrollView contentContainerStyle={styles.container}>
+        <Setting
+          label="Allow Notification"
```
fshabanov:
```suggestion
          label="Allow Notifications"
```

mobile/src/screens/profile/profile.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const handleCategoryPress = (): void => {
+    navigation.navigate(RootScreenName.SETTINGS);
+  };
+
+  return (
+    <LinearGradient>
+      <Header isArrowVisible />
+      <View style={styles.container}>
+        <Card
+          iconSourceSvg={<Icon name={'bell'} color={AppColor.WHITE} />}
```
fshabanov:
```suggestion
          iconSourceSvg={<Icon name='bell' color={AppColor.WHITE} />}
```

fshabanov (changes requested):

OlesiaTr (dismissed):

mobile/src/libs/components/icon-button/icon-button.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+const IconButton: React.FC<Properties> = ({ onPress, iconSourceSvg }) => {
+  return (
+    <TouchableOpacity onPress={onPress} style={styles.buttonContainer}>
+      {iconSourceSvg}
+    </TouchableOpacity>
+  );
+};
```
v0ldemar01: Can we use only Button component which contain only icon without label?

mobile/src/libs/maps/icon-name-to-icon.map.ts
```diff
@@ -0,0 +1,20 @@
...
+import BellIcon from '#assets/img/icons/bell.svg';
+import ChatIcon from '#assets/img/icons/chat.svg';
+import HomeIcon from '#assets/img/icons/home.svg';
+import PlusIcon from '#assets/img/icons/plus.svg';
+import UserIcon from '#assets/img/icons/user.svg';
+import { type IconName } from '#libs/types/icon-name.type';
+
+type SVGIcon = React.FC<React.SVGProps<SVGSVGElement>>;
+
+const iconNameToIcon: Record<IconName, SVGIcon> = {
```
v0ldemar01: it would be better to refer into icon package

v0ldemar01 (changes requested):

mobile/src/libs/maps/icon-name-to-icon.map.ts
iamAlinaaa: @v0ldemar01 please could you explain what do you mean ?) we need to have the same structure with frontend, thats why i created it like this and content of this file working good for now, as before i had different errors

mobile/src/libs/components/icon-button/icon-button.tsx
iamAlinaaa: @v0ldemar01 thank you for noticing)

mobile/src/libs/maps/icon-name-to-icon.map.ts
v0ldemar01: If we have an icon component in libs/components/icon so this map only linked to that components, so we don't need to expose as common map

PR #166, opened 2023-09-03
cp-52: Add user roles table

OlesiaTr:
- Add user_roles table according to db schema in readme
- Add role_id column to user
- Already existed and new users should have role 'user'

I don't know how correctly this is implemented, but I was getting all sorts of errors without inserting ROLES. This is the only solution I found working for me. 
user_roles2
user_roles1


backend/src/db/migrations/20230826133720_add_role_id_to_users_table.ts
```diff
@@ -0,0 +1,29 @@
...
+};
+
+function up(knex: Knex): Promise<void> {
+  return knex.schema.alterTable(TABLE_NAME, (table) => {
+    table
+      .integer(REFERENCE_COLUMN)
+      .unsigned()
+      .notNullable()
+      .defaultTo(ROLE_MAPPING.USER);
+    table.foreign(REFERENCE_COLUMN).references('id').inTable('user_roles');
```
fshabanov: Create enums and use 'id' and 'user_roles' from there

backend/src/db/migrations/20230826083159_add_user_roles_table.ts
```diff
@@ -0,0 +1,44 @@
...
+    key: 'chatbot',
+  },
+  { id: 2, name: 'User', key: 'user' },
+];
+
+async function up(knex: Knex): Promise<void> {
+  await knex.schema.createTable(TABLE_NAME, (table) => {
+    table.increments(ColumnName.ID).primary();
+    table.string(ColumnName.NAME).notNullable();
+    table.string(ColumnName.KEY).checkIn(['chatbot', 'user']).notNullable();
```
liza-veis: we don't need this
```suggestion
```

backend/src/db/migrations/20230826083159_add_user_roles_table.ts
```diff
@@ -0,0 +1,44 @@
...
+  ID: 'id',
+  NAME: 'name',
+  KEY: 'key',
+  CREATED_AT: 'created_at',
+  UPDATED_AT: 'updated_at',
+} as const;
+
+const ROLES = [
+  {
+    id: 1,
```
liza-veis: remove id, it will be added automatically
```suggestion
```

backend/src/db/migrations/20230826133720_add_role_id_to_users_table.ts
```diff
@@ -0,0 +1,29 @@
+import { type Knex } from 'knex';
+
+const TABLE_NAME = 'users';
+
+const REFERENCE_COLUMN = 'role_id';
+
+const ROLE_MAPPING = {
+  CHATBOT: 1,
+  USER: 2,
+};
```
liza-veis: we need to find role by key, not by id
```suggestion
const USER_ROLE_KEY = 'user'
```

backend/src/db/migrations/20230826083159_add_user_roles_table.ts
fshabanov: Create an enum with user roles and use `table.enum(ColumnName.KEY, Object.values(YOUR_ENUM))`
Do the same for Name column

fshabanov (changes requested):

backend/src/db/migrations/20230826133720_add_role_id_to_users_table.ts
liza-veis:
```suggestion
      .notNullable()
      .references(ColumnName.ID)
      .inTable(FOREIGN_TABLE_NAME);
      .defaultTo(userRoleId);
```

liza-veis (changes requested):

OlenaYatseniuk (dismissed):
With the changes above it will look good.

backend/src/db/migrations/20230826133720_add_role_id_to_users_table.ts
OlesiaTr: Done!

backend/src/db/migrations/20230826083159_add_user_roles_table.ts
OlesiaTr: Do you mean checkIn or the whole line?

backend/src/db/migrations/20230826083159_add_user_roles_table.ts
OlesiaTr: Removed it!

backend/src/db/migrations/20230826083159_add_user_roles_table.ts
OlesiaTr: Done!

backend/src/db/migrations/20230826133720_add_role_id_to_users_table.ts
OlesiaTr: I think its done?:))

backend/src/db/migrations/20230826133720_add_role_id_to_users_table.ts
OlesiaTr: Changed it!

backend/src/db/migrations/20230826083159_add_user_roles_table.ts
liza-veis: checkIn

backend/src/db/migrations/20230826083159_add_user_roles_table.ts
```diff
@@ -0,0 +1,53 @@
...
+const USER_NAMES = {
+  CHATBOT: 'Chatbot',
+  USER: 'User',
+} as const;
+
+const USER_ROLES = {
+  CHATBOT: 'chatbot',
+  USER: 'user',
+} as const;
+
```
fshabanov:
```suggestion
const ROLE_NAME = {
  CHATBOT: 'Chatbot',
  USER: 'User',
} as const;

const ROLE_KEY = {
  CHATBOT: 'chatbot',
  USER: 'user',
} as const;

```

fshabanov (changes requested):

backend/src/db/migrations/20230826083159_add_user_roles_table.ts
```diff
@@ -0,0 +1,53 @@
...
+    key: 'chatbot',
+  },
+  { name: 'User', key: 'user' },
+];
+
+async function up(knex: Knex): Promise<void> {
+  await knex.schema.createTable(TABLE_NAME, (table) => {
+    table.increments(COLUMN_NAME.ID).primary();
+    table.enum(COLUMN_NAME.NAME, Object.values(USER_NAMES));
+    table.enum(COLUMN_NAME.KEY, Object.values(USER_ROLES));
```
liza-veis: must be strings and not enums, there is no point in enum because we don't have predefined values and all values will be unique. This works, for example, if we want to set statuses and we know that we will only have three statuses, which can be repeated in some rows. But in this case, if we want to add a new role, it will never satisfy this enum and we will need to always expand it

liza-veis (changes requested):

Vika030718 (dismissed):

ossavemu (dismissed):

OlenaYatseniuk (dismissed):

backend/src/db/migrations/20230826083159_add_user_roles_table.ts
OlesiaTr: yes, ok, I returned it back

backend/src/db/migrations/20230826083159_add_user_roles_table.ts
OlesiaTr: Due to the comment from Liza, I deleted them!

backend/src/db/migrations/20230826083159_add_user_roles_table.ts
OlesiaTr: ok, done!

backend/src/db/migrations/20230826133720_add_role_id_to_users_table.ts
```diff
@@ -0,0 +1,42 @@
...
+
+type UserRow = {
+  id: number;
+};
+
+async function up(knex: Knex): Promise<void> {
+  const userRole: UserRow = (await knex(TABLE_NAME.USER_ROLES)
+    .where({ key: USER_ROLE_KEY })
+    .select(COLUMN_NAME.ID)
+    .first()) as UserRow;
```
v0ldemar01: Just include UserRow as a generic to .select<> without type assertion

v0ldemar01 (changes requested):

backend/src/db/migrations/20230826133720_add_role_id_to_users_table.ts
OlesiaTr: @v0ldemar01  then userRole might be of type undefined, what am I supposed to return in that case?

v0ldemar01: Let's proceed with such approach:
```js
const userRole = await knex(TABLE_NAME.USER_ROLES)
    .select<UserRow>(COLUMN_NAME.ID)
    .where({ key: USER_ROLE_KEY })
    .first();

  await knex.schema.alterTable(TABLE_NAME.USERS, (table) => {
    table
      .integer(COLUMN_NAME.ROLE_ID)
      .unsigned()
      .notNullable()
      .references(COLUMN_NAME.ID)
      .inTable(TABLE_NAME.USER_ROLES)
      .defaultTo((userRole as UserRow).id);
  });
```

v0ldemar01 (changes requested):

backend/src/db/migrations/20230826133720_add_role_id_to_users_table.ts
OlesiaTr: Yes, ok, changed it

v0ldemar01 (dismissed):
LGTM

backend/src/db/migrations/20230826133720_add_role_id_to_users_table.ts
```diff
@@ -0,0 +1,42 @@
...
+} as const;
+
+const COLUMN_NAME = {
+  ROLE_ID: 'role_id',
+  ID: 'id',
+} as const;
+
+const USER_ROLE_KEY = 'user';
+
+type UserRow = {
```
liza-veis:
```suggestion
type UserRole = {
```

liza-veis (changes requested):

backend/src/db/migrations/20230826133720_add_role_id_to_users_table.ts
OlesiaTr: Changed it!

v0ldemar01 (approved):

liza-veis (approved):

Nanaier (approved):

PR #169, opened 2023-09-03
cp-163: Add hover effect over buttons on the Survey page

OlenaYatseniuk:
- Added hover styling to checkbox and button on the Survey page

image
image


liza-veis:
can we just add a background color similar to selected items without borders?

OlenaYatseniuk:
> can we just add a background color similar to selected items without borders?

Yes, if you want, I can change it. For button and checkbox or only checkbox?

OlenaYatseniuk:
image


> can we just add a background color similar to selected items without borders?
Done. In the checkbox, the first item - is checked so I left a blue color to have differences with the last which only is hovered. Do you like it?
image
image

liza-veis:
Yes, looks good

frontend/src/assets/css/variables.scss
```diff
@@ -18,4 +19,6 @@
   --blue-alpha-70: #a6d4ffb2;
   --blue-alpha-30: #527deb4d;
   --gray-alpha-50: #302e3680;
+  --secondary-button-shadow: 0px 2px 4px -2px #002c840f,
```
liza-veis: this is too specific, define it where it's used

frontend/src/libs/components/checkbox/styles.module.scss
```diff
@@ -11,6 +11,11 @@
   cursor: pointer;
 }
 
+.container:hover {
+  // border: 1px solid var(--blue-700)
```
liza-veis:
```suggestion
```

liza-veis (changes requested):

fshabanov (approved):

liza-veis (approved):

PR #175, opened 2023-09-04
cp-174: Add arrow body style eslint rule with accomodation of existing changes

v0ldemar01:


fshabanov (approved):

HIMDanny (approved):
Good job!) But what is the reason for this?

v0ldemar01:
> Good job!) But what is the reason for this?

The prettier is moving function block or something like this into a new line but without readness, so it had better handle it via eslint

OlenaYatseniuk (approved):

HIMDanny:
> The prettier is moving function block or something like this into a new line but without readness, so it had better handle it via eslint

got it. Thanks for the clarification!

PR #178, opened 2023-09-04
cp-177: User Password data is returned when user sends GET authenticated-user request

HIMDanny:
- authenticated-user request now returns a user without password data

## Screenshots

image


backend/src/packages/auth/auth.service.ts
```diff
@@ -75,8 +75,11 @@ class AuthService {
     };
   }
 
-  public getAuthenticatedUser(id: number): Promise<UserAuthResponseDto | null> {
-    return this.userService.findById(id);
+  public async getAuthenticatedUser(id: number): Promise<UserAuthResponseDto> {
+    const user = (await this.userService.findById(id)) as UserAuthResponseDto;
```
OlenaYatseniuk: const user = (await this.userService.findById(id)) as UserAuthResponseDto;

What if this.userService.findById(id) returns null? Will it work as expected?

HIMDanny: I think yes. To get this response we first need to sign in. If sign-in is successful we will have this user in the db. If the opposite, then it would be the error about a nonauthorized user. So in this case `this.userService.findById(id)` wouldn't return null.

OlenaYatseniuk: @HIMDanny Good explanation!

liza-veis: we need to add modificator to the getById query not to return password from db

liza-veis: this relies on business logic too much, it's better not to expect only user here, but null also

liza-veis (changes requested):

backend/src/packages/auth/auth.service.ts
HIMDanny: ok, then I'll change it

HIMDanny: got it, will change it

liza-veis (approved):

ossavemu (approved):

iamAlinaaa (approved):

OlenaYatseniuk (approved):

v0ldemar01 (approved):

PR #205, opened 2023-09-06
cp-113: Add Mobile Profile Settings

iamAlinaaa:
1. Recreated branch due to bug that caused app to throw user out .
2. Created Common checkbox component and Setting component
3. Added profile button to header https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/issues/54
4. Improvements to reusable header
5. When user presses profile button, they should be redirected to profile screen with only one item for now (Notifications and Reminders)
6. After user presses "Notifications and Reminders" button, they should be redirected to settings page (leave only Allow Notifications setting for now)
7. Added navigation to previous screen

Снимок экрана 2023-09-02 191653
Снимок экрана 2023-09-04 123205
Снимок экрана 2023-09-04 123213


mobile/src/libs/components/button/button.tsx
```diff
@@ -17,12 +18,17 @@ const Button: React.FC<Properties> = ({
...
   type = 'solid',
+  iconSourceSvg,
 }) => {
   const handleOnPress = (): void => {
     onPress();
   };
 
-  return (
+  return iconSourceSvg ? (
+    <Pressable onPress={onPress} style={styles.iconButtonContainer}>
```
Nanaier:
```suggestion
    <Pressable onPress={handleOnPress} style={styles.iconButtonContainer}>
```

mobile/src/libs/components/card/card.tsx
```diff
@@ -2,27 +2,42 @@ import React from 'react';
...
 
-  return (
+  return iconSourceSvg ? (
+    <Pressable onPress={handlePress}>
+      <View style={styles.container}>
+        <View style={styles.iconContainer}>{iconSourceSvg}</View>
+        <Text style={styles.title}>{title}</Text>
+      </View>
+    </Pressable>
+  ) : (
```
Nanaier: Is this block really needed? you are reusing most of the styles so you can probably move the ternary operator inside the card block

iamAlinaaa: @Nanaier yeah it is needed , because this block has different structure compared to another one. and i tried to make it as you say, but layout looked badly

mobile/src/libs/enums/navigation/root-screen-name.enum.ts
```diff
@@ -3,6 +3,8 @@ const RootScreenName = {
   SIGN_IN: 'Sign In',
   SIGN_UP: 'Sign Up',
   SURVEY: 'Survey',
+  SETTINGS: 'Settings',
+  PROFILE: 'My Profile',
```
Nanaier: it is either `MY_PROFILE: 'My Profile'` or `PROFILE: 'Profile'`

mobile/src/libs/components/switch/switch.tsx
```diff
@@ -0,0 +1,41 @@
...
+  control: FormControl<T, null>;
+  name: FormFieldPath<T>;
+};
+
+const Switch = <T extends FormFieldValues>({
+  control,
+  name,
+}: Properties<T>): JSX.Element => {
+  const { field } = useFormController({ name, control });
+  const { onChange, value } = field;
```
dmytro-kucherenko:
```suggestion
  const { field: { onChange, value }} = useFormController({ name, control });
```

mobile/src/libs/components/button/button.tsx
```diff
@@ -6,8 +6,9 @@ import { Text } from '#libs/components/components';
 import { styles } from './styles';
 
 type Properties = {
-  label: string;
+  label?: string;
   onPress: () => void;
+  iconSourceSvg?: JSX.Element;
```
dmytro-kucherenko: I think `ReactNode` is more priority type, if there'll be no errors. Also it's better to name it like `iconSource` or `icon` without extra type

mobile/src/libs/components/button/button.tsx
```diff
@@ -17,12 +18,17 @@ const Button: React.FC<Properties> = ({
   onPress,
   isDisabled = false,
   type = 'solid',
+  iconSourceSvg,
 }) => {
   const handleOnPress = (): void => {
     onPress();
   };
```
fshabanov: This is an unnecessary function, please, remove it

fshabanov (changes requested):

mobile/src/libs/components/button/button.tsx
iamAlinaaa: @Dmytro-Kucherenko changed to ReactNode, but what about naming, i think this one that i use is more descriptive, to understand that its svg, because we are using different types here in code

mobile/src/libs/components/button/button.tsx
```diff
@@ -1,13 +1,14 @@
...
 
 import { Text } from '#libs/components/components';
 
 import { styles } from './styles';
 
 type Properties = {
-  label: string;
+  label?: string;
   onPress: () => void;
+  iconSourceSvg?: ReactNode;
```
liza-veis:
```suggestion
  iconName?: IconName;
```

mobile/src/libs/components/button/button.tsx
```diff
@@ -17,20 +18,21 @@ const Button: React.FC<Properties> = ({
...
+  iconSourceSvg,
 }) => {
-  const handleOnPress = (): void => {
-    onPress();
-  };
-
-  return (
+  return iconSourceSvg ? (
+    <Pressable onPress={onPress} style={styles.iconButtonContainer}>
+      {iconSourceSvg}
```
liza-veis: Icon should be inside button, also use Icon component here

mobile/src/libs/components/card/card.tsx
```diff
@@ -2,27 +2,42 @@ import React from 'react';
...
+  View,
+} from '#libs/components/components';
 
 import { DEFAULT_NUMBER_OF_LINES } from './libs/constants';
 import { styles } from './styles';
 
 type Properties = {
   title: string;
   image?: ImageSourcePropType;
+  iconSourceSvg?: JSX.Element;
```
liza-veis:
```suggestion
  iconName: IconName
```

mobile/src/libs/components/header/header.tsx
```diff
@@ -1,32 +1,73 @@
...
-import { BackButton, Badge } from './components/components';
+import { Badge } from './components/components';
 import { DEFAULT_BADGE_COUNT } from './libs/constants';
 import { styles } from './styles';
 
 type Properties = {
   title?: string;
   isArrowVisible?: boolean;
   badgeCount?: number;
+  isSettingsVisible?: boolean;
```
liza-veis: rename settigns to profile everywhere

mobile/src/libs/components/icon/maps/icon-name-to-icon.map.ts
```diff
@@ -0,0 +1,20 @@
+import BackArrowIcon from '#assets/img/icons/back-arrow.svg';
```
liza-veis: should be in components/icon/libs/maps/

mobile/src/libs/components/switch/switch.tsx
```diff
@@ -0,0 +1,42 @@
...
+const Switch = <T extends FormFieldValues>({
+  control,
+  name,
+}: Properties<T>): JSX.Element => {
+  const {
+    field: { onChange, value },
+  } = useFormController({ name, control });
+
+  return (
+    <View>
```
liza-veis: Do we need view around it?

mobile/src/libs/enums/navigation/root-screen-name.enum.ts
```diff
@@ -3,6 +3,8 @@ const RootScreenName = {
   SIGN_IN: 'Sign In',
   SIGN_UP: 'Sign Up',
   SURVEY: 'Survey',
+  SETTINGS: 'Settings',
+  MY_PROFILE: 'My Profile',
```
liza-veis: Profile should be a separate stack screen with home screen and notifications screen
```suggestion
  PROFILE: 'Profile',
```

liza-veis (changes requested):

mobile/src/libs/components/button/button.tsx
iamAlinaaa: @Liza-Veis im passing icon component inside button like on image) if i create it inside button component i will need to pass not only name but also color for icon as props to button. should i still do it ?
Снимок экрана 2023-09-07 131510

mobile/src/libs/enums/navigation/root-screen-name.enum.ts
iamAlinaaa: @Liza-Veis could you please explain, should i create separate navigation stack and add there profile and notifications (notifications like on image ?) ? because we have Home like bottom tab and whole main component includes all bottom tabs. im confused with all similar names 😅
Снимок экрана 2023-09-07 141205

OlenaYatseniuk (dismissed):
Looks good to me)

mobile/src/libs/enums/navigation/root-screen-name.enum.ts
liza-veis: no, notifications is the screen where you will get after pressing first item (notifications and reminders ) on user profile, we don't need to add one more button to header

mobile/src/libs/components/button/button.tsx
liza-veis: let's create another component for it like IconButton

liza-veis: because when we pass icon to button component we expect it to be near label inside button

iamAlinaaa: @Liza-Veis i had icon button component, @v0ldemar01 said its better approach to put icon in a button . Ill create icon button again then and make props for icon and for color f this svg icon )

mobile/src/screens/profile-settings/profile-settings.tsx
```diff
@@ -0,0 +1,30 @@
...
+  ScrollView,
+  Switch,
+} from '#libs/components/components';
+import { useAppForm } from '#libs/hooks/hooks';
+
+import { Setting } from './components/components';
+import { NOTIFICATION_SETTINGS_DEFAULT_VALUE } from './libs/constants';
+import { styles } from './styles';
+
+const ProfileSettings: React.FC = () => {
```
liza-veis:
```suggestion
const NotificationSettings: React.FC = () => {
```

mobile/src/libs/components/card/card.tsx
liza-veis:
```suggestion
  iconName?: IconName
```

mobile/src/libs/components/card/card.tsx
liza-veis: remove this condition, render either icon or this image, depending on it, not the whole block

liza-veis (changes requested):

liza-veis (dismissed):

BrianSammit (dismissed):

mobile/src/screens/notification-settings/libs/constants.ts
```diff
@@ -0,0 +1,5 @@
+const NOTIFICATION_SETTINGS_DEFAULT_VALUE = {
+  notification: true,
```
what1s1ove: boolean variables shout start with "yes" prefix
```suggestion
  hasNotification: true,
```

mobile/src/screens/notification-settings/libs/constants.ts
```diff
@@ -0,0 +1,5 @@
...
+};
```
what1s1ove:
```suggestion
} as const;
```
always add `as const` for  constants

liza-veis (approved):

PR #227, opened 2023-09-08
cp-187: User Not Found Appearance

OlesiaTr:
The description of the bug: User not found error appears after logging in with valid credentials and refreshing the page for people, who had unchanged database since day 1.

After changing the modifier: The page refreshed, and no errors appeared.


v0ldemar01 (dismissed):

backend/src/packages/users/user.model.ts
```diff
@@ -31,11 +31,11 @@ class UserModel extends AbstractModel {
     return {
       withoutPassword(builder): QueryBuilder<UserModel> {
         return builder.select(
```
liza-veis, line 33: what is returned after this query? don't we need to add other fields like surveyIsCopmleted etc

liza-veis (changes requested):

backend/src/packages/users/user.model.ts
liza-veis: or are they added automatically?

OlesiaTr: They are added in the findById method in the user.repo

dmytro-kucherenko: As I understand, it means that it can be quite misunderstanding that we don't return other required fields there. We got them indeed, but can't see that in code

OlesiaTr: well, modifier was added to make sure that password information wouldn't be seen where it;s not needed at all. But overall in the findById method the Graph is joined with the user details, where we get all extra information

```
 const user = await this.userModel
      .query()
      .modify('withoutPassword')
      .withGraphJoined(UsersRelation.DETAILS)
      .findById(id)
      .castTo<UserCommonQueryResponse | undefined>()
      .execute();
```

backend/src/packages/users/user.model.ts
```diff
@@ -31,11 +31,11 @@ class UserModel extends AbstractModel {
...
-          'details.id',
+          'users.id',
           'email',
           'details.fullName',
-          'details.created_at',
-          'details.updated_at',
+          'users.created_at',
+          'users.updated_at',
```
liza-veis: Let's add users even if fields don't conflict, and also we don't need to specify details fields
```suggestion
          'users.id',
          'users.email',
          'users.created_at',
          'users.updated_at',
```

liza-veis (changes requested):

backend/src/packages/users/user.model.ts
Ekit18: But we could also do smth like this instead of modifier
      .modify((builder) => {
        void builder.where('users.id', id).select('users.id', 'users.email', 'users.createdAt', 'users.updatedAt', 'details.fullName').first();
      })

OlesiaTr: I changed to users, checked that its returning data properly

liza-veis (approved):

Issue #190, opened 2023-09-05
Chat functionality


liza-veis:
- [x] when user selects a chat, all messages of this chat should be loaded
- [x] when user sends message to the chat it should be saved on the backend

Depends on create chat functionality #189

Xxandder:
@Liza-Veis Hello, I want to clarify some implementing aspects not to redo everything in future, because I am not sure in exact way of implementing it. Am I right that I need to create separate package for messages with model/repository/entity/cervice/controller for backend?  I ask because my task is related to cp-189 and there is needed to create backend part for chats. And the same question for storage. Should I create separate slice or maybe we can create slice with something like this: store = {chats[], currentChatMessages[]}

Xxandder:
@Liza-Veis Hello. I have two questions to clarify. 
1. Do I need to throw error when chat is not exists when user tries to create message with wrong chatId? The same for the senderId. 
2. What is the max length of the message?

liza-veis:
1. Yes, if course, we should handle all cases
2. let's have 4000 chars limit

liza-veis:
> Am I right that I need to create separate package for messages with model/repository/entity/cervice/controller for backend

yes, we need separate chat and chat-message packages, but we will have only one controller in chat package

> Should I create separate slice or maybe we can create slice with something like this: store = {chats[], currentChatMessages[]}

the second option - one slice with chat and currentChat properties

Xxandder:
@Liza-Veis Do we really need "name" column in chat_messages table?


![Image](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/111736110/102f07ca-0552-4a97-823b-79739446f47e)

liza-veis:
No, we don't, remove it

Xxandder:
@Liza-Veis Could you please clarify the next question? 
I created types for backend requests and responses for chat-messages. But I am not sure which fields are enough for project functionality. So in chat layout we can use only message text and senderId and it will be enough for current tasks. But maybe in future we will to edd extra functionality where chatId is needed. So can we use this type?:
```
type ChatMessageGetAllItemResponseDto = {
  id: number;
  senderId: number;
  message: string;
};
```

liza-veis:
created and updatedAt should be returned as well

liza-veis:
and let's return chatId too, so that we can detect the chat by message on frontend

Xxandder:
ok, got it

[...]

PR #85, opened 2023-08-26
cp-65: Add Chat Layout

ossavemu:
The layout of the Chat was added and the requirements requested in the issue were met. The attached images are from the project after building. The following command was used to verify the results: `npm run build:frontend && npm run preview -w frontend`

Captura de pantalla 2023-08-26 153645
Captura de pantalla 2023-08-26 153433
DRFREUD

PR #236, opened 2023-09-09
cp-189: Create chat functionality

HIMDanny:
### Backend
- Chat entity/model/repository/service/controller
- Methods for finding chats of the authenticated user and creating a chat
- Swagger docs to chat endpoints

### Frontend
- Chat API
- Chat slice
- Get all chats and create chat actions
- Load and render chats in the sidebar chat list
- A new empty chat is opened when the user presses the create chat button
- A new chat is created and it appears in the chat list when the user sends the first message

Note: I used a mocked chat name with numbers for now because we cannot have two chats with the same name in the DB.

#### Get all chats

image

#### Create a chat

image

#### New empty chat

image

#### Newly created chat added to the chat list

image


backend/src/db/migrations/20230906103635_add_members_column_to_chats_table.ts
```diff
@@ -0,0 +1,27 @@
+import { type Knex } from 'knex';
+
+const TABLE_NAME = 'chats';
+
+const ColumnName = {
+  MEMBERS: 'members',
```
liza-veis: let's go with another approach and create users_to_chats table instead of members column, so that we can automatically delete user from chat if user was deleted

liza-veis (changes requested):

shared/src/packages/chats/libs/validation-schemas/create-chat.validation-schema.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { ChatValidationMessage } from '../enums/enums.js';
+import { type ChatRequestDto } from '../types/types.js';
+
+const createChat = joi.object<ChatRequestDto, true>({
+  name: joi
+    .string()
+    .trim()
+    .required()
+    .messages({ 'any.required': ChatValidationMessage.NAME_REQUIRED }),
```
dmytro-kucherenko: Also 'string.required' and 'string.empty' is needed in order to avoid situations with entered space e.t.c.

dmytro-kucherenko (commented):
Looks good for now

shared/src/packages/chats/libs/validation-schemas/create-chat.validation-schema.ts
HIMDanny: thanks, will add this

HIMDanny: do we also need `string.required` when we have `any.required`?

backend/src/db/migrations/20230906103635_add_members_column_to_chats_table.ts
HIMDanny: So should this table look like this?
```mermaid
erDiagram
users_to_chats {
  int id PK
  dateTime created_at
  dateTime updated_at
  int user_id FK
  int chat_id FK
}
```

And I should delete the migration for members column then, right?

liza-veis: yes

shared/src/packages/chats/libs/validation-schemas/create-chat.validation-schema.ts
liza-veis: there is no `string.required` according to joi docs

dmytro-kucherenko: Yeah, I missed that, I got confused with another message type. So adding only `string.empty` will be enough

HIMDanny: done

readme.md
```diff
@@ -81,6 +81,14 @@ erDiagram
       dateTime updated_at
       varchar name
       int topic_id FK
+      array members
```
liza-veis:
```suggestion
```

liza-veis (changes requested):
Nice job 🙌 waiting for #85 to be merged

shared/src/packages/chats/chats.ts
```diff
@@ -0,0 +1,8 @@
+export { ChatValidationMessage } from './libs/enums/enums.js';
+export { ChatsApiPath } from './libs/enums/enums.js';
```
OlesiaTr, line 1:
```suggestion
export { ChatValidationMessage , ChatsApiPath } from './libs/enums/enums.js';
```

HIMDanny: good point! Missed that

OlesiaTr (dismissed):

OlenaYatseniuk (dismissed):

frontend/src/slices/chats/chats.slice.ts
```diff
@@ -0,0 +1,54 @@
...
+import { type ValueOf } from '#libs/types/types.js';
+import { type ChatGetAllItemResponseDto } from '#packages/chats/chats.js';
+
+import { createChat, getAllChats } from './actions.js';
+
+type State = {
+  chats: ChatGetAllItemResponseDto[];
+  chatsDataStatus: ValueOf<typeof DataStatus>;
+  createChatDataStatus: ValueOf<typeof DataStatus>;
+};
```
danilittaa: I see that, for example, for 'journal.slice', we name properties in such a way: '
image
 (with the word 'entries' at the end). Perhaps we should use the same approach everywhere, what do you think?

liza-veis: for journalEntries we name it like this because it's a name of our entity (journal_entry, like note), while journal is a feature name. Here chat is a name of our entity and a feature name as well

HIMDanny:
@Liza-Veis, should I handle the error about the same chat name, or it is unnecessary because it will be always a newly generated name from AI?

liza-veis:
it's ok to have chat's with the same name

backend/src/packages/chats/chat.entity.ts
```diff
@@ -0,0 +1,98 @@
...
+  public static initializeNew({
+    name,
+  }: {
+    name: string;
+    members: UserToChatQueryResponse[];
+  }): ChatEntity {
+    return new ChatEntity({
+      id: null,
+      name,
+      members: [],
```
liza-veis, line 62:
```suggestion
      members,
```

HIMDanny:
oh, then I need to change db constraint

HIMDanny:
do I need to create a migration for this or change directly?

frontend/src/pages/chat/components/chat-footer/chat-footer.tsx
```diff
@@ -19,6 +19,10 @@ const ChatFooter: React.FC<Properties> = ({ onSend }) => {
 
   const onSubmit = useCallback(
     ({ text }: ChatInputValue): void => {
+      if (!text.trim()) {
+        return;
+      }
```
liza-veis: this should be handled by validation schema, we will add this in chat messaging issue #190
```suggestion
```

shared/src/packages/chats/libs/validation-schemas/create-chat.validation-schema.ts
```diff
@@ -0,0 +1,13 @@
...
+
+import { ChatValidationMessage } from '../enums/enums.js';
+import { type ChatRequestDto } from '../types/types.js';
+
+const createChat = joi.object<ChatRequestDto, true>({
+  name: joi.string().trim().required().messages({
+    'any.required': ChatValidationMessage.NAME_REQUIRED,
+    'string.empty': ChatValidationMessage.NAME_REQUIRED,
+  }),
+});
```
liza-veis: we will generate name on the backend side, so for now just set default name with chat id in backend service like `Chat #12`
```suggestion
```

frontend/src/index.tsx
```diff
@@ -83,6 +84,16 @@ createRoot(document.querySelector('#root') as HTMLElement).render(
...
+                children: [
+                  {
+                    path: AppRoute.CHATS_$ID,
+                    element: (
+                      <ProtectedRoute>
+                        <ChatMessages />
+                      </ProtectedRoute>
+                    ),
+                  },
+                ],
```
liza-veis: We should do something similar to sign-in/sign-up flow, use one component and in it detect by parameters is it a new chat or existed one
```suggestion
                children: [
                  {
                    path: AppRoute.CHATS_$ID,
                    element: (
                      <ProtectedRoute>
                        <Chat />
                      </ProtectedRoute>
                    ),
                  },
                ],
```

liza-veis (changes requested):

backend/src/packages/chats/chat.service.ts
```diff
@@ -0,0 +1,52 @@
...
+      items: items.map((item) => {
+        return item.toObject();
+      }),
+    };
+  }
+
+  public async create(
+    payload: CreateChatPayload,
+  ): Promise<ChatGetAllItemResponseDto> {
+    const item = await this.chatRepository.create(payload);
```
liza-veis, line 38: don't pass payload directly, use initializeNew
```suggestion
    const item = await this.chatRepository.create(payload);
```

liza-veis (changes requested):

shared/src/packages/chats/libs/validation-schemas/create-chat.validation-schema.ts
HIMDanny: no problem

frontend/src/pages/chat/components/chat-footer/chat-footer.tsx
HIMDanny: I understand, just made it for temporal use, till it is not implemented. Ok, I will remove it

shared/src/packages/chats/libs/validation-schemas/create-chat.validation-schema.ts
HIMDanny: I thought we would generate a name by sending another request and it would depend on the user's message

backend/src/packages/chats/chat.entity.ts
HIMDanny: In this case, I should type members as `UserToChatQueryResponse[] | number[]`, right?

backend/src/packages/chats/chat.service.ts
HIMDanny: If so, I would need to type members as `UserToChatQueryResponse[] | number[]`, or create a `users-to-chats.entity` to initialize it first and then pass it to the chat entity. What do you think will be the preferable way?

v0ldemar01:
> do I need to create a migration for this or change directly?

Any db changes should be updated only with migration

backend/src/packages/chats/chat.entity.ts
v0ldemar01: It could be a map from one type to another

HIMDanny: I'm afraid I didn't understand what you meant

liza-veis: let's leave empty array, but remove `members: UserToChatQueryResponse[];` from initializeNew parameters. We will pass members separately

HIMDanny: yeah, I left it there by chance

backend/src/packages/chats/chat.service.ts
HIMDanny: @Liza-Veis, as we decided to remove members from the new initializer, do I still need to create here a chat entity, and pass it to the create method with members argument separately?

shared/src/packages/chats/libs/validation-schemas/create-chat.validation-schema.ts
liza-veis: i think we will send first message with the chat creation, because we can't create messages before chat is created and we can't add chat name without message

backend/src/packages/chats/chat.service.ts
liza-veis: yes, use initializeNew to pass basic entry and pass members as a second parameter

shared/src/packages/chats/libs/validation-schemas/create-chat.validation-schema.ts
HIMDanny: how about creating an endpoint like `/first-message` to handle this situation?

liza-veis: this doesn't fix the situation, because we can't create a message without chat, so I have two options:

1. send request to generate name without creating a message, then request to create chat, then request to create message - 3 requests needed (will take too much time for the first message)
2. send request to create chat with the first message, generate name on backend, create chat, create message - everything in one request

HIMDanny: I see. Then we would choose the second option, and in this case, I need to add a message to the create chat request body

HIMDanny: and creating a message would be on @Xxandder in this issue #190

OlesiaTr (dismissed):

frontend/src/index.tsx
HIMDanny: but why we can't simply use outlet to render messages?

frontend/src/pages/chat/libs/constants/constants.ts
```diff
@@ -1,44 +1,40 @@
-import { UserRole } from '#libs/enums/enums.js';
-import { generateUUID } from '#libs/helpers/helpers.js';
-import { type ChatMessage } from '#libs/types/types.js';
+import { type ChatMessageGetAllItemResponseDto } from '#packages/chat-messages/chat-messages.js';
+
+const EMPTY_ARRAY_LENGTH = 0;
```
dmytro-kucherenko: It's common constant, lets store it in `shared/src/libs/constants` folder. At now this approach isn't implemented on development branch, but hopefully it will after some mergings

dmytro-kucherenko (commented):
Nice, only small suggestion)

frontend/src/pages/chat/libs/constants/constants.ts
HIMDanny: good point! Changed it

v0ldemar01 (dismissed):

frontend/src/pages/chat/chat-messages.tsx
```diff
@@ -0,0 +1,27 @@
...
+  });
+
+  return (
+    <>
+      {MOCK_MESSAGES.map((item) => {
+        return (
+          <ChatMessage
+            key={item.id}
+            message={item.message}
+            isSender={item.senderId === authenticatedUser?.id}
```
fshabanov: We will always have a user in this page, cast the type

frontend/src/pages/chat/components/chat-footer/chat-footer.tsx
```diff
@@ -8,7 +8,7 @@ import { type ChatInputValue } from '#pages/chat/libs/types/types.js';
 import styles from './styles.module.scss';
 
 type Properties = {
-  onSend: ({ text }: { text: string }) => void;
+  onSend: ({ message }: { message: string }) => void;
```
fshabanov:
```suggestion
  onSend: ({ message }: ChatInputValue) => void;
```

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
```diff
@@ -1,34 +1,60 @@
...
-        {MOCK_MESSAGES.map((item) => {
-          return (
-            <ChatMessage
-              key={item.id}
-              item={item}
-              isSender={item.sender === UserRole.USER}
-            />
-          );
-        })}
+        {!!id &&
```
fshabanov:
```suggestion
        {Boolean(id) &&
```

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
```diff
@@ -1,34 +1,60 @@
...
+    ({ chats, auth }) => {
+      return {
+        currentChatMessages: chats.currentChatMessages,
+        authenticatedUser: auth.authenticatedUser,
+      };
+    },
+  );
+
+  const handleSend = useCallback(
+    ({ message }: { message: string }): void => {
```
fshabanov:
```suggestion
    ({ message }: ChatInputValue): void => {
```

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
```diff
@@ -1,34 +1,60 @@
...
-            />
-          );
-        })}
+        {!!id &&
+          MOCK_MESSAGES.map((item) => {
+            return (
+              <ChatMessage
+                key={item.id}
+                message={item.message}
+                isSender={item.senderId === authenticatedUser?.id}
```
fshabanov: Cast the type

frontend/src/pages/chat/components/chat-sidebar/chat-sidebar.tsx
```diff
@@ -40,7 +40,9 @@ const ChatSidebar: React.FC = () => {
           </span>
         </div>
         <div className={styles['plus']}>
-          <Icon name={'plus'} color={IconColor.BLUE} />
+          <Link to="/chats">
+            <Icon name={'plus'} color={IconColor.BLUE} />
```
fshabanov:
```suggestion
            <Icon name='plus' color={IconColor.BLUE} />
```

frontend/src/pages/chat/components/chat-sidebar/chat-sidebar.tsx
```diff
@@ -49,14 +51,20 @@ const ChatSidebar: React.FC = () => {
...
-                onClick={handleSelectChat(filteredChat.id)}
-                isActive={selectedChatId === filteredChat.id}
-                key={filteredChat.id}
-              />
+              <Link key={filteredChat.id} to={chatLink}>
+                <Card
+                  title={filteredChat.name}
+                  imageUrl={cardPlaceholder}
+                  onClick={handleSelectChat}
+                  isActive={false}
```
fshabanov: false is by default

fshabanov (changes requested):

frontend/src/pages/chat/chat-messages.tsx
HIMDanny: good point

frontend/src/pages/chat/components/chat-sidebar/chat-sidebar.tsx
HIMDanny: you are right, missed that

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
```diff
@@ -1,34 +1,62 @@
...
+              <ChatMessage
+                key={item.id}
+                message={item.message}
+                isSender={item.senderId === authenticatedUser.id}
+              />
+            );
+          })}
       </div>
-      <ChatFooter onSend={handleSend} />
+      <ChatFooter key={id} onSend={handleSend} />
```
fshabanov:
```suggestion
      <ChatFooter onSend={handleSend} />
```
It is not in a loop

fshabanov (changes requested):

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
HIMDanny: You are right. It was made for clearing a chat input after changing routes

backend/src/db/migrations/20230914101723_remove_unique_constraint_in_chats_name_column.ts
```diff
@@ -0,0 +1,21 @@
...
+
+function up(knex: Knex): Promise<void> {
+  return knex.schema.table(TABLE_NAME, (table) => {
+    table.dropUnique([ColumnName.NAME]);
+  });
+}
+
+function down(knex: Knex): Promise<void> {
+  return knex.schema.table(TABLE_NAME, (table) => {
+    table.unique([ColumnName.NAME]);
```
liza-veis, line 46: we need to handle case here when we have non unique items, maybe query non unique items and update their names with index

liza-veis (changes requested):

backend/src/db/migrations/20230914101723_remove_unique_constraint_in_chats_name_column.ts
HIMDanny: ok, I will handle it

backend/src/packages/chats/chat.service.ts
liza-veis, line 38: use initializeNew to pass basic entry and pass members as a second parameter

backend/src/db/migrations/20230914101723_remove_unique_constraint_in_chats_name_column.ts
```diff
@@ -0,0 +1,50 @@
+import { type Knex } from 'knex';
+
+type Chat = {
+  id: number;
+  name: string;
+};
+
+const NAME_INDEX_ADDITION = 1;
```
liza-veis:
```suggestion
const NAME_INDEX_INCREMENT_NUMBER = 1;
```

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
liza-veis:
```suggestion
      <ChatFooter onSend={handleSend} />
```

shared/src/packages/chats/libs/types/chat-get-all-item-response-dto.type.ts
```diff
@@ -0,0 +1,9 @@
+import { type UserToChatResponseDto } from './user-to-chat-response-dto.type.js';
+
+type ChatGetAllItemResponseDto = {
+  id: number;
+  name: string;
+  members: UserToChatResponseDto[];
```
liza-veis: define properties directly here and remove UserToChatResponseDto
```suggestion
  members: {...}[];
```

liza-veis (changes requested):

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
fshabanov: I think it is better to use `reset` inside of `useEffect` for this purpose

backend/src/packages/chats/chat.service.ts
HIMDanny: I am passing it, but as an object with two params, because `create` method in `Service` needs us to pass only one parameter

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
HIMDanny: It was made on purpose for clearing a chat input after changing routes. If I remove this then the chat input will not clear its value. I am updating the chat footer every time we change the route. It is just a simple method of doing so)

HIMDanny: here is the link to official documentation about this: https://react.dev/reference/react/useState#resetting-state-with-a-key

frontend/src/index.tsx
```diff
@@ -58,14 +58,6 @@ createRoot(document.querySelector('#root') as HTMLElement).render(
...
                 ),
               },
-              {
-                path: AppRoute.SIGN_IN,
-                element: <Auth />,
-              },
-              {
-                path: AppRoute.SIGN_UP,
-                element: <Auth />,
-              },
```
dmytro-kucherenko, line 68: Maybe I'm missing smth, but why are these routes just deleted?

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
```diff
@@ -1,32 +1,60 @@
...
-        {MOCK_MESSAGES.map((item) => {
-          return (
-            <ChatMessage
-              key={item.id}
-              item={item}
-              isSender={item.sender === UserRole.USER}
-            />
-          );
-        })}
+        {Boolean(id) &&
```
dmytro-kucherenko: I think creating `hasId` variable will make it more clear

HIMDanny:
@Liza-Veis, @fshabanov, I have changed the logic of resetting the chat message input

frontend/src/index.tsx
HIMDanny: I have deleted the duplicates

dmytro-kucherenko: Got it

Xxandder (approved):
Looks good to me)

fshabanov (approved):

liza-veis (approved):

PR #239, opened 2023-09-10
cp-234: Add Journal Notes Loading

danilittaa:
- added journal icon to tab navigation;
- journal slice
- journal screen

image


mobile/src/slices/journal/journal.slice.ts
```diff
@@ -0,0 +1,59 @@
...
+    builder.addCase(createJournalEntry.pending, (state) => {
+      state.createJournalEntryDataStatus = DataStatus.PENDING;
+    });
+    builder.addCase(createJournalEntry.fulfilled, (state, action) => {
+      state.allJournalEntries.push(action.payload);
+      state.createJournalEntryDataStatus = DataStatus.FULFILLED;
+    });
+    builder.addCase(createJournalEntry.rejected, (state) => {
+      state.createJournalEntryDataStatus = DataStatus.REJECTED;
+    });
```
liza-veis: we don't need to add this in this PR
```suggestion
```

liza-veis (changes requested):

liza-veis (dismissed):

mobile/src/packages/journal/journal-api.ts
```diff
@@ -0,0 +1,33 @@
+import { APIPath } from '#libs/enums/enums';
+import { BaseHttpApi } from '#libs/packages/api/api';
+import { type HTTP } from '#libs/packages/http/http';
+import { type Storage } from '#libs/packages/storage/storage';
+
+import { JournalApiPath } from './libs/enums/enums';
+import { type JournalEntryGetAllResponseDto } from './libs/types/types.js';
```
fshabanov:
```suggestion
import { type JournalEntryGetAllResponseDto } from './libs/types/types';
```

mobile/src/packages/journal/journal-api.ts
```diff
@@ -0,0 +1,33 @@
...
+  http: HTTP;
+  storage: Storage;
+};
+
+class JournalApi extends BaseHttpApi {
+  public constructor({ baseUrl, http, storage }: Constructor) {
+    super({ path: APIPath.JOURNAL, baseUrl, http, storage });
+  }
+
+  public async getAllJournalEntriers(): Promise<JournalEntryGetAllResponseDto> {
```
fshabanov:
```suggestion
  public async getAllJournalEntries(): Promise<JournalEntryGetAllResponseDto> {
```

fshabanov (changes requested):

v0ldemar01 (dismissed):
LGTM, please follow Farid's comments

liza-veis (approved):

mobile/src/packages/journal/libs/enums/enums.ts
```diff
@@ -0,0 +1 @@
+export { JournalApiPath } from 'shared/build/index.js';
```
liza-veis, line 1: .js extension won't work on mobile
```suggestion
export { JournalApiPath } from 'shared/build/index';
```

mobile/src/packages/journal/libs/types/types.ts
```diff
@@ -0,0 +1,4 @@
+export {
+  type JournalEntryGetAllItemResponseDto,
+  type JournalEntryGetAllResponseDto,
+} from 'shared/build/index.js';
```
liza-veis, line 4:
```suggestion
} from 'shared/build/index';
```

liza-veis (changes requested):

liza-veis:
And also make padding inside the scrollView, not outside it, so that we see scrollbar at the edge of the screen

mobile/src/screens/journal/journal.tsx
```diff
@@ -0,0 +1,89 @@
...
+  useCallback,
+  useEffect,
+  useNavigation,
+  useSearch,
+} from '#libs/hooks/hooks';
+import { actions as journalActions } from '#slices/journal/journal';
+
+import { styles } from './styles';
+
+const mockedCount = 12;
```
dmytro-kucherenko: Lets use constants file or at least leave blank line after)

dmytro-kucherenko (commented):
Nice job!

mobile/src/screens/journal/journal.tsx
danilittaa: I deleted that line and now use the actual value. Thank you)

mobile/src/packages/journal/libs/enums/enums.ts
danilittaa: But we have shared/build/index.JS in many files in the mobile folder (screenshot 1), and without .js, I get an error (screenshot 2)
image
image

liza-veis: ok, I will need to check this, let's leave for now

HIMDanny (dismissed):

mobile/src/libs/components/card/styles.ts
```diff
@@ -16,6 +16,7 @@ const styles = StyleSheet.create({
     paddingRight: 6,
     paddingVertical: 3,
     shadowColor: AppColor.PURPLE_100,
+    marginHorizontal: 12,
```
liza-veis: we don't want to add margin inside card, this should be done by padding inside scroll view

mobile/src/screens/journal/journal.tsx
```diff
@@ -51,7 +51,7 @@ const Journal: React.FC = () => {
         );
       },
     });
-  }, [navigation, route.name]);
+  }, [navigation, route.name, allJournalEntries]);
```
liza-veis: let's move allJournalEntries.length to variable and use it inside useEffect

liza-veis (changes requested):

iamAlinaaa (commented):
looks good , and maybe where the plus icon we should write + add new note instead of + add new journal ?

danilittaa:
> looks good , and maybe where the plus icon we should write + add new note instead of + add new journal ?

@Liza-Veis, what do you think about this?

liza-veis:
Yes, it should be add new note

mobile/src/screens/journal/journal.tsx
```diff
@@ -0,0 +1,90 @@
+import React from 'react';
+
+import PlusIcon from '#assets/img/icons/plus.svg';
```
v0ldemar01: Let's put it as existing icons

mobile/src/screens/journal/journal.tsx
```diff
@@ -0,0 +1,90 @@
...
+              <Card
+                title={item.title}
+                onPress={handleSelectJournal}
+                key={item.id}
+              />
+            );
+          })}
+        </ScrollView>
+        <View style={styles.linkWrapper}>
+          <Link
```
v0ldemar01, line 77: Maybe it would be Button?

v0ldemar01 (changes requested):

mobile/src/screens/journal/journal.tsx
```diff
@@ -0,0 +1,90 @@
...
+  );
+
+  const handleSelectJournal = useCallback(() => {
+    // TODO: Implement actual functionality for the onPress event
+  }, []);
+
+  useEffect(() => {
+    navigation.setOptions({
+      header: () => {
+        return <Header title={route.name} badgeCount={badgeCount} />;
```
fshabanov:
```suggestion
        return <Header badgeCount={badgeCount} />;
```
Title is route name by default. You can also remove everything related to `useAppRoute` from this file

fshabanov (changes requested):

mobile/src/screens/journal/journal.tsx
danilittaa: Can I change the type of the label from string to React.ReactNode in the Button component to pass an icon? I could pass the icon separately as a prop, but I already have such an implementation in my other task 'Add Meditation Player Layout', with completely different logic. Or should I go back to using a different Button component in 'Add Meditation Player Layout' as it was before and then pass the icon as a prop here?

liza-veis: let's leave it for now, maybe we will change this in future

mobile/src/screens/journal/journal.tsx
v0ldemar01: Not resolved

mobile/src/libs/components/icon/icon.tsx
```diff
@@ -7,12 +7,13 @@ import { iconNameToIcon } from './libs/maps/maps';
 type Properties = {
   name: IconName;
   color?: string;
+  styles?: React.CSSProperties;
```
liza-veis: we should not pass styles to our components, if you need to add margins/paddings or positioning then wrap it in View, in case you need specific styles - pass specific properties
```suggestion
```

mobile/src/screens/journal/journal.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  useEffect(() => {
+    void dispatch(journalActions.getAllJournalEntries());
+  }, [dispatch]);
+
+  return (
+    <LinearGradient>
+      <View style={styles.container}>
+        <InputSearch
+          placeholder="Search journal"
```
liza-veis:
```suggestion
          placeholder="Search note"
```

liza-veis (changes requested):

liza-veis (approved):

fshabanov (approved):

v0ldemar01 (approved):

PR #266, opened 2023-09-13
cp-70: Fixed issues with file uploading functionality

Nanaier:
Fixed the following issues:

- When the user tries to add a file > than 10MB:
 Now the code is 400.

Screenshot 2023-09-13 at 16 02 03

- When the user adds a file in any other format than .png or .jpg
 Now the code is 400.

Screenshot 2023-09-13 at 16 02 49

- Updated swagger to include id, url, and contentType and dates

Screenshot 2023-09-13 at 16 32 05

- Updated data returned from the backend on file entry creation

Screenshot 2023-09-13 at 16 32 27


backend/src/packages/files/file.service.ts
```diff
@@ -62,15 +62,14 @@ class FileService implements Service {
...
-    await this.fileRepository.create(
+    const file = await this.fileRepository.create(
       FileEntity.initializeNew({
         url,
         contentType: payload.contentType,
       }),
     );
 
-    return { url: presignedUrl };
+    return { id: file.toObject().id, url, contentType: payload.contentType };
```
liza-veis:
```suggestion
    return file.toObject()
```

liza-veis (changes requested):

backend/src/libs/plugins/file-upload/file-upload.plugin.ts
```diff
@@ -15,7 +15,9 @@ const fileUpload = fp<Options>((fastify, { extensions }, done) => {
 
   fastify.addHook(ControllerHook.PRE_VALIDATION, async (request) => {
     if (!request.isMultipart()) {
-      return;
+      throw new FileError({
```
dmytro-kucherenko: You're throwing file error, that as I understand has internal server error status. Should this error description also be in swagger documentation?

Nanaier: hey - I checked once again and catching this error breaks down the logic behind generating swagger documentation, so this check had to be removed

dmytro-kucherenko: Nice, got it

dmytro-kucherenko (approved):

liza-veis (approved):

PR #288, opened 2023-09-14
cp-246: White circles around sidebar options don't have all the same size

juanc-escobar:
I fixed these issues:
The icons didn't have the same size.  
The link was not the same shape as the button container.  
The icon container didn't have a fixed width.

image

image


OlesiaTr (dismissed):

frontend/src/libs/components/icon/icon.tsx
```diff
@@ -5,12 +5,19 @@ import { iconNameToIcon } from './libs/maps/icon-name-to-icon.map.js';
 type Properties = {
   name: IconName;
   color?: string;
+  width?: string | number;
+  height?: string | number;
```
dmytro-kucherenko: It's better to make these properties `number` only and then convert them to `string` and add `px`

frontend/src/libs/components/icon/icon.tsx
```diff
@@ -5,12 +5,25 @@ import { iconNameToIcon } from './libs/maps/icon-name-to-icon.map.js';
...
+  width,
+  height,
+}) => {
   const SelectedIcon = iconNameToIcon[name];
 
-  return <SelectedIcon style={{ color }} />;
+  const svgProperties = {
+    color,
+    width: width ? `${width}px` : undefined,
+    height: height ? `${height}px` : undefined,
```
dmytro-kucherenko: We're trying not to set undefined, maybe null if it's allowed or empty single quote

liza-veis: can't we pass width and height as attributes instead of styles?

liza-veis (changes requested):

OlesiaTr (approved):

dmytro-kucherenko (approved):

v0ldemar01 (approved):

Ekit18 (approved):

fshabanov (approved):

iamAlinaaa (approved):

liza-veis (approved):

PR #292, opened 2023-09-14
cp-170: Add Meditation Player Layout

danilittaa:
- fixed all comment in the issue

- added 'react-native-keep-awake' library;
- removed button toggling during 30-second rewind forward or backward;
- changed some styles;
- changed color when user presses button: 
image


mobile/src/libs/components/button/button.tsx
```diff
@@ -22,7 +25,30 @@ const Button: React.FC<Properties> = ({
...
+  const [isPressed, setIsPressed] = useState(false);
+
+  let color;
+  if (isPressed) {
+    color = AppColor.BLUE_200;
+  } else {
+    color = isRounded ? AppColor.BLUE_200 : AppColor.GRAY_400;
+  }
+
+  const handleOnPress = (): void => {
```
dmytro-kucherenko:
```suggestion
  const handlePress = (): void => {
```

mobile/src/libs/components/player/player.tsx
```diff
@@ -48,15 +45,25 @@ const Player: React.FC<Properties> = ({ setCurrentTrack }) => {
...
+      KeepAwake.deactivate();
     };
+  }, []);
 
+  useEffect(() => {
+    const addPlaylist = async (): Promise<void> => {
+      void player.setPlaylist(MOCKED_PLAYLIST);
+      setPlaybackState(await player.getState());
+    };
     void addPlaylist();
```
dmytro-kucherenko: Is it necessary to create function to call it once in the next line?

dmytro-kucherenko (dismissed):

mobile/src/libs/components/button/button.tsx
```diff
@@ -22,7 +25,30 @@ const Button: React.FC<Properties> = ({
   type = 'solid',
   isRounded,
   iconName,
+  isVisuallyCentered,
 }) => {
+  const [isPressed, setIsPressed] = useState(false);
+
+  let color;
```
v0ldemar01: Create a separate helper outside for this purpose

mobile/src/libs/components/button/button.tsx
```diff
@@ -22,7 +25,30 @@ const Button: React.FC<Properties> = ({
...
+  }
+
+  const handlePress = (): void => {
+    setIsPressed(true);
+    onPress();
+  };
+
+  useEffect(() => {
+    if (isPressed) {
+      setTimeout(() => {
```
v0ldemar01: What about clearTimeout?

v0ldemar01 (changes requested):

mobile/src/libs/hooks/use-press-state/use-press-state.hook.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const usePressState = ({ onPress, isRounded }: Properties): PressState => {
+  const [isPressed, setIsPressed] = useState(false);
+
+  let color;
+  if (isPressed) {
+    color = AppColor.BLUE_200;
+  } else {
+    color = isRounded ? AppColor.BLUE_200 : AppColor.GRAY_400;
+  }
```
v0ldemar01:
```suggestion
  const color = isPressed || isRounded ? AppColor.BLUE_200 : AppColor.GRAY_400;
```

v0ldemar01 (changes requested):

mobile/src/libs/hooks/use-press-state/use-press-state.hook.ts
```diff
@@ -0,0 +1,46 @@
+import { AppColor } from '#libs/enums/enums';
```
liza-veis: remove this hook, use android ripple instead

liza-veis (changes requested):

OlesiaTr (dismissed):

mobile/src/libs/hooks/use-press-state/use-press-state.hook.ts
danilittaa: I've tried using this before, but the problem is that it specifically changes the background, not the color. However, I just realized that I can use the activeOpacity property, and it looks something like this:

https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/108797614/34b707f7-e0bf-4c92-8f88-201b01603323


 What do you think about it?

liza-veis: I like it

v0ldemar01 (approved):

iamAlinaaa (approved):
good job)

liza-veis (approved):

PR #293, opened 2023-09-14
cp-194: Add and edit journal note functionality

OlenaYatseniuk:
Changes:
- add creating new note by click on plus button in journal sidebar
- when user clicks on note item in sidebar, this note is loaded
- content is saved to backend automatically with debounce - in two seconds after user stops typing
-  if the page reloads up-to-date note content is sent to backend 
- all user input data is sanitized from both sides - backend and frontend
- create new migration to add column with userId in journals_entry table
- create update and create methods for journalEntry controller, service and repository
- add redux actions for update note

https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/102380564/90f8ccdc-aa2a-4cad-bd09-811c8cea738c


backend/src/packages/journal-entries/journal-entry.repository.ts
```diff
@@ -37,28 +58,43 @@ class JournalEntryRepository implements Repository {
...
+      .patchAndFetchById(id, { title, text })
+      .castTo<JournalEntryCommonQueryResponse>();
+
+    return JournalEntryEntity.initialize({
+      id: updatedJournalEntry.id,
+      userId: updatedJournalEntry.userId,
+      text: updatedJournalEntry.text,
+      title: updatedJournalEntry.title,
+      createdAt: new Date(updatedJournalEntry.createdAt),
+      updatedAt: new Date(),
```
v0ldemar01:
```suggestion
      updatedAt: new Date(updatedJournalEntry.updatedAt),
```

frontend/package.json
```diff
@@ -27,11 +27,14 @@
     "react-hook-form": "7.45.4",
     "react-redux": "8.1.2",
     "react-router-dom": "6.15.0",
-    "react-toastify": "9.1.3"
+    "react-toastify": "9.1.3",
+    "sanitize-html": "2.11.0",
```
v0ldemar01: It should be only in shared

frontend/src/libs/components/note/note.tsx
```diff
@@ -0,0 +1,193 @@
...
+  const { value: titleValue, onChange: onTitleChange } = titleField;
+  const { value: textValue, onChange: onTextChange } = textField;
+
+  const dispatch = useAppDispatch();
+  const { id } = useParams();
+
+  const titleReference = useRef<HTMLDivElement | null>(null);
+  const textReference = useRef<HTMLDivElement | null>(null);
+
+  const [isFirstRender, setIsFirstRender] = useState(true);
```
v0ldemar01: Cannot figure out the purpose for it

frontend/src/libs/components/note/note.tsx
```diff
@@ -0,0 +1,193 @@
...
+    }, [onTitleChange]);
+
+  const handleTextChange: React.FormEventHandler<HTMLDivElement> =
+    useCallback(() => {
+      if (textReference.current) {
+        const newText = textReference.current.textContent ?? '';
+        onTextChange(newText);
+        changeCursorPosition(cursorPosition);
+      }
+    }, [onTextChange]);
```
v0ldemar01: Create a common core helper

frontend/src/libs/components/note/note.tsx
```diff
@@ -0,0 +1,193 @@
...
+      handleSaveNoteWithDebounce.clear();
+    };
+  }, [handleSaveNote, handleSaveNoteWithDebounce, textValue, titleValue]);
+
+  if (
+    journalEntriesDataStatus === DataStatus.IDLE ||
+    journalEntriesDataStatus === DataStatus.PENDING
+  ) {
+    return <Loader />;
+  }
```
v0ldemar01: Maybe thats logic for parent component

frontend/src/pages/journal/components/journal-sidebar/journal-sidebar.tsx
```diff
@@ -1,25 +1,41 @@
...
 
 const JournalSidebar: React.FC = () => {
   const dispatch = useAppDispatch();
-  const { allJournalEntries, selectedJournalEntry } = useAppSelector(
-    ({ journal }) => {
+
+  const { allJournalEntries, selectedJournalEntry, userId } = useAppSelector(
+    ({ journal, auth }) => {
       return {
+        userId: auth.authenticatedUser?.id,
```
v0ldemar01: Use type assertion here

frontend/src/pages/journal/components/journal-sidebar/journal-sidebar.tsx
```diff
@@ -49,6 +72,11 @@ const JournalSidebar: React.FC = () => {
                 onClick={handleSelectJournalEntry(journalEntry.id)}
                 isActive={selectedJournalEntry?.id === journalEntry.id}
                 key={journalEntry.id}
+                linkTo={
+                  `${AppRoute.JOURNAL}/${journalEntry.id}` as ValueOf<
```
v0ldemar01: Create a full path and with replace put the value here. The example of Chat functionality's PR

frontend/src/slices/journal/actions.ts
```diff
@@ -1,22 +1,27 @@
...
 
 const createJournalEntry = createAsyncThunk<
   JournalEntryGetAllItemResponseDto,
   JournalEntryCreateRequestDto,
   AsyncThunkConfig
->(`${sliceName}/create-journal-entry`, async (payload, { extra }) => {
+>(`${sliceName}/create-journal-entry`, async (payload, { extra, dispatch }) => {
   const { journalApi } = extra;
+  const journalEntry = await journalApi.createJournalEntry(payload);
+  dispatch(appActions.navigate(`${AppRoute.JOURNAL}/${journalEntry.id}`));
```
v0ldemar01: Let's use existing logic of navigation

v0ldemar01 (changes requested):

backend/src/libs/types/service.type.ts
```diff
@@ -1,10 +1,10 @@
 type Service<T = unknown> = {
-  find(): Promise<T>;
+  find(id: number): Promise<T>;
   findAll(): Promise<{
     items: T[];
   }>;
-  create(payload: unknown): Promise<T>;
-  update(): Promise<T>;
+  create(payload: unknown, id?: number): Promise<T>;
+  update(id: number, userId: number, payload: unknown): Promise<T>;
```
HIMDanny: I guess these params are not so generic. Maybe let's use one payload as an object to pass needed params?

backend/src/packages/journal-entries/journal-entry.controller.ts
```diff
@@ -109,23 +178,135 @@ class JournalEntryController extends BaseController {
...
    *      responses:
+   *        201:
+   *          description: Successful operation
+   *          content:
+   *            application/json:
+   *              schema:
+   *                type: object
+   *                properties:
+   *                  journalEntry:
+   *                    $ref: '#/components/schemas/Journal Entry'
```
HIMDanny: Does the response actually look like this? If I'm not mistaken it is something like `{ items: [{ ... }] }`

backend/src/packages/journal-entries/journal-entry.controller.ts
```diff
@@ -109,23 +178,135 @@ class JournalEntryController extends BaseController {
...
+   *      responses:
+   *        201:
+   *          description: Successful operation
+   *          content:
+   *            application/json:
+   *              schema:
+   *                type: object
+   *                properties:
+   *                  journalEntry:
+   *                    $ref: '#/components/schemas/Journal Entry'
```
HIMDanny: I guess it should look like this. Please correct me if it shouldn't
```suggestion
   *              schema:
   *                $ref: '#/components/schemas/Journal Entry'
```

backend/src/packages/journal-entries/journal-entry.repository.ts
```diff
@@ -14,20 +14,41 @@ class JournalEntryRepository implements Repository {
     this.journalEntryModel = journalEntryModel;
   }
 
-  public find(): ReturnType<Repository['find']> {
-    return Promise.resolve(null);
+  public async find(id: number): Promise<JournalEntryEntity | null> {
+    const journalEntry = await this.journalEntryModel
+      .query()
+      .findById(id)
+      .execute();
```
HIMDanny:
```suggestion
```

backend/src/packages/journal-entries/journal-entry.service.ts
```diff
@@ -31,19 +50,57 @@ class JournalEntryService implements Service {
...
-      JournalEntryEntity.initializeNew({
-        title: payload.title,
-        text: payload.text,
+      JournalEntryEntity.initialize({
+        id: null,
+        userId,
+        createdAt: null,
+        updatedAt: null,
+        title: sanitizeInput(payload.title),
+        text: sanitizeInput(payload.text),
```
HIMDanny: why we don't use `initializeNew` here? You already added the `useId` as a property there

frontend/src/libs/components/card/card.tsx
```diff
@@ -20,37 +21,44 @@ const Card: React.FC<Properties> = ({
...
-          {hasIcon && (
-            <div className={styles['icon-background']}>
-              <Icon name={iconName as IconName} color={iconColor} />
-            </div>
-          )}
-        </div>
-      )}
-      <div className={styles['title']}>{title}</div>
-    </button>
+    <Link to={linkTo ?? '/'}>
```
HIMDanny: why don't just wrap a card into a link in a parent component?

frontend/src/libs/components/note/note.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+type Properties = {
+  className: string;
+};
+
+const Note: React.FC<Properties> = ({ className }) => {
+  const { userId, selectedJournalEntry, journalEntriesDataStatus } =
+    useAppSelector(({ auth, journal }) => {
+      return {
+        userId: auth.authenticatedUser?.id,
```
HIMDanny: we can assert type here. The user will see the note only if he's authenticated

frontend/src/libs/components/card/card.tsx
OlenaYatseniuk: Yeah, I rewrote it similar to your component.

backend/src/packages/journal-entries/journal-entry.service.ts
OlenaYatseniuk: You are right, thanks for noticing.

backend/src/packages/journal-entries/journal-entry.controller.ts
OlenaYatseniuk: You are right, thanks for noticing.

backend/src/packages/journal-entries/journal-entry.service.ts
```diff
@@ -29,21 +49,56 @@ class JournalEntryService implements Service {
...
+    const user = await userService.findById(userId as number);
+
+    if (!user) {
+      throw new UsersError({
+        status: HTTPCode.NOT_FOUND,
+        message: ExceptionMessage.USER_NOT_FOUND,
+      });
+    }
+
+    const newJournalEntry = await this.journalEntryRepository.update(
```
liza-veis:
```suggestion
    const item = await this.journalEntryRepository.update(
```

frontend/package.json
```diff
@@ -27,7 +27,8 @@
     "react-hook-form": "7.45.4",
     "react-redux": "8.1.2",
     "react-router-dom": "6.15.0",
-    "react-toastify": "9.1.3"
+    "react-toastify": "9.1.3",
+    "vite-plugin-svgr": "3.2.0"
```
liza-veis: we already have it in dev dependencies
```suggestion
```

frontend/src/libs/components/note/libs/constants.ts
```diff
@@ -0,0 +1,8 @@
+const NOTE_TIMEOUT = 2000;
```
liza-veis:
```suggestion
const SAVE_NOTE_TIMEOUT = 2000;
```

frontend/src/libs/components/note/note.tsx
```diff
@@ -0,0 +1,175 @@
...
+  const dispatch = useAppDispatch();
+  const { id } = useParams();
+
+  const titleReference = useRef<HTMLDivElement | null>(null);
+  const textReference = useRef<HTMLDivElement | null>(null);
+  const cursorPosition = useRef<number | null>(null);
+
+  const handleSaveNote = useCallback(
+    (data: NoteContent) => {
+      if (data.title && data.text && id) {
```
liza-veis: we need to save it even there is no title or text, also let's pass id in params, not to check it's presence twice
```suggestion
  const handleSaveNote = useCallback(
    (id: number, data: NoteContent) => {
```

shared/src/packages/journal/libs/validation-schemas/create-journal-entry.validation-schema.ts
```diff
@@ -0,0 +1,30 @@
+import joi from 'joi';
+
+import {
+  JournalEntryValidationMessage,
+  JourneyEntryValidationRule,
+} from '../enums/enums.js';
+import { type JournalEntryCreateRequestDto } from '../types/types.js';
+
+const createJournalEntryValidationSchema = joi.object<
```
liza-veis:
```suggestion
const createJournalEntry = joi.object<
```
and then  `createJournalEntry as createJournalEntryValidationSchema` in re-export

shared/src/packages/journal/libs/types/journal-entry-update-request-dto.type.ts
```diff
@@ -0,0 +1,10 @@
+type JournalEntryUpdateRequestDto = {
+  id: number;
+  userId?: number;
```
liza-veis: we can't update user id
```suggestion
```

shared/src/packages/journal/libs/types/journal-entry-update-request-dto.type.ts
```diff
@@ -0,0 +1,10 @@
...
+  body: {
+    title: string;
+    text: string;
+  };
```
liza-veis: let's remove body
```suggestion
    title: string;
    text: string;
```

shared/src/libs/helpers/sanitize-input/sanitize-input.helper.ts
```diff
@@ -0,0 +1,17 @@
+import sanitizeHtml, { type IOptions } from 'sanitize-html';
+
+const sanitizeInput = (input: string, options?: IOptions): string => {
+  const charactersToRemove = /[&<>]/g;
+  const cleanedInput = input.replaceAll(charactersToRemove, '');
```
liza-veis: won't sanitizeHtml remove all html characters?
```suggestion
```

shared/src/libs/helpers/sanitize-input/sanitize-input.helper.ts
```diff
@@ -0,0 +1,17 @@
+import sanitizeHtml, { type IOptions } from 'sanitize-html';
```
liza-veis:
```suggestion
import sanitizeHtml, { type IOptions as SanitizeHtmlOptions } from 'sanitize-html';
```

frontend/src/pages/journal/journal.tsx
```diff
@@ -1,17 +1,40 @@
...
+  useState,
+} from '#libs/hooks/hooks.js';
 
 import { JournalSidebar } from './components/journal-sidebar/journal-sidebar.js';
 import styles from './styles.module.scss';
 
 const Journal: React.FC = () => {
+  const { id } = useParams();
+  const dispatch = useAppDispatch();
+  const [isNoteVisible, setIsNoteVisible] = useState(false);
```
liza-veis: also useEffect can be removed
```suggestion
  const hasNoteId = Boolean(id) && selectedJournalEntry;
```

frontend/src/pages/journal/components/journal-sidebar/journal-sidebar.tsx
```diff
@@ -24,15 +29,28 @@ const JournalSidebar: React.FC<Properties> = ({
...
       return {
+        userId: (auth.authenticatedUser as UserAuthResponseDto).id,
         allJournalEntries: journal.allJournalEntries,
         selectedJournalEntry: journal.selectedJournalEntry,
       };
     },
   );
 
+  const handlePlusButtonClick = useCallback(() => {
+    if (userId) {
```
Liza-Veis, line 44: we don't need this check
```suggestion
```

frontend/src/libs/hooks/use-app-form/use-app-form.hook.ts
```diff
@@ -23,7 +24,9 @@ type ReturnValue<T extends FieldValues = FieldValues> = {
   control: Control<T, null>;
   errors: FieldErrors<T>;
   isValid: boolean;
+  isDirty: boolean;
   handleSubmit: UseFormHandleSubmit<T>;
+  watch: WatchInternal<T>;
```
liza-veis: looks like we don't use it
```suggestion
```

frontend/src/libs/components/note/note.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+import { NOTE_TIMEOUT } from './libs/constants.js';
+import { type NoteContent } from './libs/types.js';
+import styles from './styles.module.scss';
+
+type Properties = {
+  className: string;
+};
+
+const Note: React.FC<Properties> = ({ className }) => {
```
liza-veis: should be in journal screen components

frontend/src/libs/components/note/note.tsx
```diff
@@ -0,0 +1,175 @@
...
+} from '#libs/hooks/hooks.js';
+import { type UserAuthResponseDto } from '#packages/users/users.js';
+import { actions as journalActions } from '#slices/journal/journal.js';
+
+import { NOTE_TIMEOUT } from './libs/constants.js';
+import { type NoteContent } from './libs/types.js';
+import styles from './styles.module.scss';
+
+type Properties = {
+  className: string;
```
liza-veis: we shouldn't pass className to custom components
```suggestion
```

frontend/src/libs/components/note/note.tsx
```diff
@@ -0,0 +1,175 @@
...
+        userId: (auth.authenticatedUser as UserAuthResponseDto).id,
+        selectedJournalEntry: journal.selectedJournalEntry,
+      };
+    },
+  );
+
+  const { isDirty, control } = useAppForm({
+    defaultValues: {
+      title: selectedJournalEntry?.title,
+      text: selectedJournalEntry?.text,
```
liza-veis: we need to render this component only when selectedJournalEntry is present
```suggestion
      title: selectedJournalEntry.title,
      text: selectedJournalEntry.text,
```

frontend/src/libs/components/note/note.tsx
```diff
@@ -0,0 +1,175 @@
...
+  useEffect(() => {
+    if (selectedJournalEntry) {
+      onTitleChange(selectedJournalEntry.title);
+      onTextChange(selectedJournalEntry.text);
+    }
+  }, [onTitleChange, onTextChange, selectedJournalEntry, isDirty]);
+
+  useEffect(() => {
+    if (titleReference.current && cursorPosition.current) {
+      getCursorPosition(titleReference.current, cursorPosition);
```
liza-veis: why do we need all this logic with getCursorPosition?

frontend/src/libs/components/note/note.tsx
```diff
@@ -0,0 +1,175 @@
...
+  useEffect(() => {
+    void dispatch(journalActions.getAllJournalEntries()).then(() => {
+      if (id) {
+        void dispatch(journalActions.setSelectedJournalEntry(Number(id)));
+      }
+    });
+  }, [dispatch, id]);
+
+  useEffect(() => {
+    const handleSaveNoteWithDebounce = debounce((data: NoteContent) => {
```
liza-veis: looks like this won't work, because we will create new debounce function on every change. but we need to call this function only after the last change. At least we need to move debounce outside useEffect to always reference to the same funciton

For example: user types something, if less than 2 seconds passed between new changes, nothing saves. Only after they stop typing (more than 2 seconds passed), request is sent

I'm not sure if we don't lose reference in debounce, looks like we should wrap it into custom hook alse 🤔

liza-veis (changes requested):

OlesiaTr (dismissed):

frontend/src/libs/components/note/note.tsx
OlenaYatseniuk: But if we send empty note data (without title and text) we might have many empty cards in the sidebar, I think it will not be OK.

shared/src/packages/journal/libs/types/journal-entry-update-request-dto.type.ts
OlenaYatseniuk: I did it because I use this type for backend and frontend (in journal API). So for frontend I don't need userId, but in backend I need it, as for create and update  methods in service I need have one payload, so I can't pass it as second argument. So what can I do with it?

shared/src/libs/helpers/sanitize-input/sanitize-input.helper.ts
OlenaYatseniuk: It doesn't remove it, just changes it for example "<" to ";" and "&" to "amp;". So what should I do with this?

frontend/src/libs/components/note/note.tsx
OlenaYatseniuk: Well, without this logic div contentEditable sets the cursor at the beginning and adds characters before the first character ( like from right to left, not as we want - from left to right). Also, we cannot put a cursor somewhere in the middle of the text if you want to edit something there.

frontend/src/libs/components/note/note.tsx
OlenaYatseniuk: But it works as expected now =). It calls the debounced function only in 2 sec when the user stops typing. When I tried to have this function outside with useCallback I had a warning that I needed to put this function without callback as its dependencies are unknown. Also, we have a similar example of debouncing usage in the search component.  You can see it in this screenshot.  So everything works as expected now.
image

frontend/src/libs/components/note/note.tsx
liza-veis: in the example [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable) it works well, or you mean something else?

frontend/src/libs/components/note/note.tsx
liza-veis: but how can this work if debounce() always returns a new function?

frontend/src/libs/helpers/get-cursor-position.helper.ts
```diff
@@ -0,0 +1,50 @@
+import {
+  EMPTY_ARRAY_LENGTH,
+  FIRST_ARRAY_INDEX,
+} from '#libs/constants/constants.js';
+
+const getCursorPosition = (
```
v0ldemar01: Get method returns a value, not mutate it

frontend/src/libs/helpers/get-cursor-position.helper.ts
```diff
@@ -0,0 +1,50 @@
...
+    const selection = window.getSelection() as Selection;
+
+    let adjustedCursorPosition = cursorPosition.current;
+
+    if (
+      adjustedCursorPosition &&
+      adjustedCursorPosition > (element.textContent as string).length
+    ) {
+      adjustedCursorPosition = (element.textContent as string).length;
+    }
```
v0ldemar01:
```suggestion
    let adjustedCursorPosition = cursorPosition.current;

    let adjustedCursorPosition = cursorPosition.current;
    if (!adjustedCursorPosition) {
      return;
    }
    adjustedCursorPosition = Math.min(adjustedCursorPosition, (element.textContent as string).length)
```

frontend/src/pages/journal/components/journal-sidebar/journal-sidebar.tsx
```diff
@@ -53,17 +71,30 @@ const JournalSidebar: React.FC<Properties> = ({
...
+          onClick={handlePlusButtonClick}
+        />
       </SidebarHeader>
       <SidebarBody>
         <div className={styles['journal-entry-list']}>
           {allJournalEntries.map((journalEntry) => {
+            const noteLink = (AppRoute.JOURNAL_ENTRY_$ID.replace(
+              ':id',
+              String(journalEntry.id),
+            ) + '?sidebarMode=hide') as typeof AppRoute.JOURNAL_ENTRY_$ID;
```
v0ldemar01: it's useless here

v0ldemar01: If you want to hide a sidebar prefer using callback for it

frontend/src/pages/journal/components/note/note.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const { isDirty, control } = useAppForm({
+    defaultValues: {
+      title: selectedJournalEntry.title,
+      text: selectedJournalEntry.text,
+    },
+    mode: 'onChange',
+  });
+
+  const { field: titleField } = useFormController({ name: 'title', control });
+  const { field: textField } = useFormController({ name: 'text', control });
```
v0ldemar01: Let's split note component, this one is responsible for many targets

frontend/src/libs/components/note/note.tsx
OlenaYatseniuk: I am not sure why I have this behavior, but it looks like this. Also there are many requests in Google with this problem.

https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/102380564/2662d08f-e365-492d-b954-f03862c546b7

shared/src/packages/journal/libs/enums/journal-entry-validation-message.enum.ts
```diff
@@ -0,0 +1,8 @@
+const JournalEntryValidationMessage = {
+  USER_ID_MUST_BE_NUMBER: 'User id must be a number.',
+  TITLE_REQUIRED: 'Title is required.',
+  TITLE_MUST_BE_LESS_THAN_255: 'Title must be less than 255 characters.',
```
v0ldemar01: Import JournalEntryValidationRule here(see thread-js)

v0ldemar01 (changes requested):

frontend/src/libs/components/note/note.tsx
OlenaYatseniuk: You can see how it works now here)
https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/102380564/c4146a19-596b-4f9c-b2c9-d2e3e9d3af6f

v0ldemar01: Side effects may work sometimes but it is not a best practise

v0ldemar01: Let's put callback into debounce without condition

v0ldemar01 (changes requested):

frontend/src/libs/components/note/note.tsx
OlenaYatseniuk: @v0ldemar01  Can you, please clarify what you mean? If like in 1 screenshot, so it sends request every 2 sec, so it doesn't work as expected. If in 2 variant I have a warning here:
image
image

frontend/src/libs/components/note/note.tsx
OlenaYatseniuk: Should I change it to regular input so as not to have this logic?

frontend/src/pages/journal/components/journal-sidebar/journal-sidebar.tsx
OlenaYatseniuk: I did it like it is in the chatSidebar component, which was merged. Also, there is a bug [https://github.com/orgs/BinaryStudioAcademy/projects/12/views/1?sliceBy%5Bvalue%5D=HIMDanny&pane=issue&itemId=38741468](bug) created for this. So it is a temporary solution.

frontend/src/pages/journal/components/note/note.tsx
OlenaYatseniuk: I am not sure that I understand what you mean. Could you explain this,please?

frontend/src/libs/components/note/note.tsx
v0ldemar01: You are receiving the error, bcos you missed to pass handleSaveNote as deps to handleSaveNoteWithDebounce

shared/src/packages/journal/libs/types/journal-entry-update-request-dto.type.ts
liza-veis: then it should be separate types

frontend/src/libs/components/note/note.tsx
OlenaYatseniuk: it doesn't help =(
image

OlenaYatseniuk: if I make it like this, I don't have an error, but I can't use handleSaveNoteWithDebounce.clear() anymore. So how could I handle it?
image

OlesiaTr (dismissed):

frontend/src/pages/journal/components/journal-sidebar/journal-sidebar.tsx
liza-veis: it was fixed, so you can change this

frontend/src/libs/components/note/note.tsx
OlenaYatseniuk: @Liza-Veis  Also, I found a problem with a cursor. After the page reloads, or first rendering the component I can't type something, because cursor jumps to the beginning instead of being in the end, and it is impossible to type. If I change note for another one, I can type and update it. And I can't figure out how to deal with this problem.

frontend/src/pages/journal/components/note/note.tsx
OlenaYatseniuk: Sorry, I don't understand how to split this logic. As for me, it looks pretty solid.

backend/src/packages/journal-entries/journal-entry.service.ts
```diff
@@ -15,12 +21,26 @@ class JournalEntryService implements Service {
     this.journalEntryRepository = journalEntryRepository;
   }
 
-  public find(): ReturnType<Service['find']> {
-    return Promise.resolve(null);
+  public async find(
```
liza-veis: will 404 error return to frontend if no item found?

backend/src/packages/journal-entries/journal-entry.service.ts
```diff
@@ -29,21 +49,57 @@ class JournalEntryService implements Service {
...
+        status: HTTPCode.NOT_FOUND,
+        message: ExceptionMessage.USER_NOT_FOUND,
+      });
+    }
+
+    const item = await this.journalEntryRepository.update(
+      JournalEntryEntity.initialize({
+        id,
+        userId,
+        createdAt: null,
```
liza-veis, line 76: check that this doesn't override createdAt and updatedAt properties

frontend/src/pages/journal/components/journal-sidebar/styles.module.scss
```diff
@@ -4,6 +4,14 @@
   align-items: center;
 }
 
+.list {
+  display: flex;
+  flex-grow: 1;
+  flex-direction: column;
+  align-items: center;
+  height: 90%;
+}
```
liza-veis: do we need this after we fixed sidebar? especially height 90%. styles should be almost the same as in chat sidebar

frontend/src/pages/journal/components/note/note.tsx
```diff
@@ -0,0 +1,158 @@
...
+  const dispatch = useAppDispatch();
+  const { id } = useParams();
+
+  const titleReference = useRef<HTMLDivElement | null>(null);
+  const textReference = useRef<HTMLDivElement | null>(null);
+  const cursorPosition = useRef<number | null>(null);
+
+  const handleSaveNote = useCallback(
+    (id: string, data: NoteContent) => {
+      if (data.text && data.title) {
```
liza-veis: let's allow to save note without text, and if there is no title let's save it as default note payload - 'New note'

frontend/src/pages/journal/journal.tsx
```diff
@@ -1,16 +1,47 @@
...
     setIsSidebarShown(true);
-  }, [setIsSidebarShown]);
+  }, [setIsSidebarShown, navigate]);
+
+  useEffect(() => {
+    void dispatch(journalActions.getAllJournalEntries()).then(() => {
+      if (id) {
+        void dispatch(journalActions.setSelectedJournalEntry(Number(id)));
+      }
+    });
```
liza-veis: remove `then` from here and dispatch the second action inside journalActions.getAllJournalEntries

frontend/src/pages/journal/journal.tsx
```diff
@@ -1,16 +1,47 @@
...
+  const dispatch = useAppDispatch();
   const { isSidebarShown, setIsSidebarShown } = useSidebarState();
 
+  const { selectedJournalEntry } = useAppSelector(({ journal }) => {
+    return {
+      selectedJournalEntry: journal.selectedJournalEntry,
+    };
+  });
+
+  const hasNoteId = Boolean(id) && selectedJournalEntry;
```
liza-veis: let's rename it to this to make more logical
```suggestion
  const hasSelectedNote = Boolean(id) && selectedJournalEntry;
```

shared/src/libs/helpers/sanitize-input/sanitize-input.helper.ts
liza-veis: it's ok if it will change tags to mnemonic, that's what we need

liza-veis (changes requested):

frontend/src/pages/journal/components/note/note.tsx
liza-veis: let's leave it as it is

backend/src/packages/journal-entries/journal-entry.service.ts
OlenaYatseniuk: Yeah, it works as expected.

frontend/src/pages/journal/components/journal-sidebar/styles.module.scss
OlenaYatseniuk: Yes, thanks for noticing, i will replace it as in chat sidebar

frontend/src/pages/journal/journal.tsx
OlenaYatseniuk: but in this way, I need to pass id to getAllJournalEntries and it will look weird as for me.

frontend/src/pages/journal/journal.tsx
OlenaYatseniuk: Done

frontend/src/pages/journal/components/note/note.tsx
OlenaYatseniuk: Done

backend/src/packages/journal-entries/journal-entry.service.ts
OlenaYatseniuk: Done

frontend/src/pages/journal/journal.tsx
liza-veis: can't we use await syntax, maybe without void?

OlenaYatseniuk: @Liza-Veis then i will have this error
image

OlenaYatseniuk: I can change it to this. Is this ok?
image

frontend/src/pages/journal/components/journal-sidebar/journal-sidebar.tsx
```diff
@@ -24,15 +30,28 @@ const JournalSidebar: React.FC<Properties> = ({
...
       return {
+        userId: (auth.authenticatedUser as UserAuthResponseDto).id,
         allJournalEntries: journal.allJournalEntries,
         selectedJournalEntry: journal.selectedJournalEntry,
       };
     },
   );
 
+  const handlePlusButtonClick = useCallback(() => {
+    if (userId) {
```
liza-veis: we don't need this check

frontend/src/pages/journal/journal.tsx
liza-veis: can't we just split it into two useEffects? looks like we don't need to await all journal entries to set selected journal entry

backend/src/db/migrations/20230909153115_add_user_id_to_journals_entries_table.ts
liza-veis, line 1: `journals_entries` -> `journal_entries` in file name

backend/src/packages/journal-entries/libs/enums/table-columns/journal-entries-table-column.ts
```diff
@@ -0,0 +1,10 @@
+const JournalsEntriesTableColumn = {
```
liza-veis:
```suggestion
const JournalEntriesTableColumn = {
```

liza-veis (changes requested):

OlenaYatseniuk:
No, we need to wait for the getAllJournalEntries fulfilled to select an item. And if it splits to 2 effect it will cause infinity sending setAllJournalEntries

frontend/src/pages/journal/journal.tsx
OlenaYatseniuk: No, we need to wait for the getAllJournalEntries fulfilled to select an item. And if it splits to 2 effect it will cause infinity sending setAllJournalEntries

liza-veis: ok, then let's use as on screenshot above

backend/src/packages/journal-entries/journal-entry.repository.ts
```diff
@@ -14,20 +14,38 @@ class JournalEntryRepository implements Repository {
...
   }
 
-  public async findAll(): Promise<JournalEntryEntity[]> {
+  public async findAllByUserId(userId: number): Promise<JournalEntryEntity[]> {
     const journalEntries = await this.journalEntryModel
       .query()
-      .select()
-      .castTo<JournalEntryCommonQueryResponse[]>()
-      .execute();
+      .where('userId', userId)
```
v0ldemar01:
```suggestion
      .where({ userId )
```

backend/src/packages/journal-entries/journal-entry.service.ts
```diff
@@ -29,21 +49,57 @@ class JournalEntryService implements Service {
...
+    userId,
+  }: CreateJournalEntryPayload): Promise<JournalEntryGetAllItemResponseDto> {
+    const user = await userService.findById(userId);
+
+    if (!user) {
+      throw new UsersError({
+        status: HTTPCode.NOT_FOUND,
+        message: ExceptionMessage.USER_NOT_FOUND,
+      });
+    }
```
v0ldemar01: As we are using auth user, the prehandler decoded token to userId and found user, is it a mandatory thing to make this one here

backend/src/packages/journal-entries/journal-entry.service.ts
```diff
@@ -29,21 +49,57 @@ class JournalEntryService implements Service {
...
-    return Promise.resolve(null);
+  public async update({
+    id,
+    userId,
+    title,
+    text,
+  }: JournalEntryUpdateRequestDto): Promise<JournalEntryGetAllItemResponseDto> {
+    const user = await userService.findById(userId);
+
+    if (!user) {
```
v0ldemar01: Same here

shared/src/packages/journal/libs/enums/journal-entry-validation-message.enum.ts
```diff
@@ -0,0 +1,8 @@
+import { JourneyEntryValidationRule } from './journal-entry-validation-rule.enum.js';
+
+const JournalEntryValidationMessage = {
+  TITLE_REQUIRED: 'Title is required.',
+  TITLE_MUST_BE_LESS_THAN_255: `Title must be less than ${JourneyEntryValidationRule.MAXIMUM_TITLE_LENGTH} characters.`,
```
v0ldemar01:
```suggestion
  TITLE_MUST_BE_LESS_THAN: `Title must be less than ${JourneyEntryValidationRule.MAXIMUM_TITLE_LENGTH} characters.`,
```

v0ldemar01 (changes requested):
LGTM, please check last comments

backend/src/packages/journal-entries/journal-entry.service.ts
OlenaYatseniuk: Ok, I will remove it.

OlenaYatseniuk:
@Liza-Veis  @v0ldemar01 I made all the changes, but, unfortunately, I found a bug with the cursor position. So when it is a new user and he creates a new note, or after the page reloading ( so it is the first rendering of the note component) it is impossible to type something to update the note because the cursor jumps from the end to beginning and it doesn't allow to type anything. But if you have another note that was updated previously you can check this note and it is possible to update every note after this changes. I tried to handle it, but I can't understand what exactly caused this issue.
 So, without cursor handling I can't type as expected as it puts the cursor at the beginning and I can write only from beginning to end ( so it is not as usual typing) and also cannot put the cursor anywhere in the middle).
 I think the problem is with react and contentEditable, because in MDN  it works okay without handling the cursor, but in our development, it doesn't. So now, I think we should decide how to fix it. Or try to handle it with the cursor deeper, or try to find react lib for contendEditable that can handle this or rewrite it to usual input. So, in all these ways I think I need someone's help to handle it quicker to close this issue ASAP.  
 

https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/102380564/b6c92101-9d40-4703-b8a9-4a6f88712fa0

v0ldemar01:
> @Liza-Veis @v0ldemar01 I made all the changes, but, unfortunately, I found a bug with the cursor position. So when it is a new user and he creates a new note, or after the page reloading ( so it is the first rendering of the note component) it is impossible to type something to update the note because the cursor jumps from the end to beginning and it doesn't allow to type anything. But if you have another note that was updated previously you can check this note and it is possible to update every note after this changes. I tried to handle it, but I can't understand what exactly caused this issue. So, without cursor handling I can't type as expected as it puts the cursor at the beginning and I can write only from beginning to end ( so it is not as usual typing) and also cannot put the cursor anywhere in the middle). I think the problem is with react and contentEditable, because in MDN it works okay without handling the cursor, but in our development, it doesn't. So now, I think we should decide how to fix it. Or try to handle it with the cursor deeper, or try to find react lib for contendEditable that can handle this or rewrite it to usual input. So, in all these ways I think I need someone's help to handle it quicker to close this issue ASAP.
> 
>  cursor-bug.mp4

Did you check react-contenteditable library?

OlenaYatseniuk:
> > @Liza-Veis @v0ldemar01 I made all the changes, but, unfortunately, I found a bug with the cursor position. So when it is a new user and he creates a new note, or after the page reloading ( so it is the first rendering of the note component) it is impossible to type something to update the note because the cursor jumps from the end to beginning and it doesn't allow to type anything. But if you have another note that was updated previously you can check this note and it is possible to update every note after this changes. I tried to handle it, but I can't understand what exactly caused this issue. So, without cursor handling I can't type as expected as it puts the cursor at the beginning and I can write only from beginning to end ( so it is not as usual typing) and also cannot put the cursor anywhere in the middle). I think the problem is with react and contentEditable, because in MDN it works okay without handling the cursor, but in our development, it doesn't. So now, I think we should decide how to fix it. Or try to handle it with the cursor deeper, or try to find react lib for contendEditable that can handle this or rewrite it to usual input. So, in all these ways I think I need someone's help to handle it quicker to close this issue ASAP.
> > cursor-bug.mp4
> 
> Did you check react-contenteditable library?

I checked it when started working on this issue. But I faced many ts problems, so that's why I decided to make it without lib. I can try to rewrite it now and check if the cursor problem isn't present there.

OlenaYatseniuk:
@v0ldemar01  Well, it can work with this lib, but with useState. Because with useAppForm hook I have problem to handle changes. And I also have this ts problem. I am not sure how to fix this.
image

v0ldemar01:
> @v0ldemar01 Well, it can work with this lib, but with useState. Because with useAppForm hook I have problem to handle changes. And I also have this ts problem. I am not sure how to fix this. image

Could you push the latest changes to github?

frontend/src/pages/journal/journal.tsx
```diff
@@ -1,16 +1,49 @@
...
+  const { selectedJournalEntry } = useAppSelector(({ journal }) => {
+    return {
+      selectedJournalEntry: journal.selectedJournalEntry,
+    };
+  });
+
+  const hasSelectedNote = Boolean(id) && selectedJournalEntry;
+
   const handleBackButtonPress = useCallback(() => {
+    navigate(AppRoute.JOURNAL);
```
v0ldemar01, line 34: We have appActions.navigate for this

v0ldemar01 (changes requested):

OlesiaTr (dismissed):

OlenaYatseniuk:
> > @v0ldemar01 Well, it can work with this lib, but with useState. Because with useAppForm hook I have problem to handle changes. And I also have this ts problem. I am not sure how to fix this. image
> 
> Could you push the latest changes to github? P.S. I didn't delete cursor helper, but I remember about it. 

@v0ldemar01, I did it with no-verify flag because I have type errors
P.S.I didn't delete the cursor helper, but I remember about it.

v0ldemar01:
> > > @v0ldemar01 Well, it can work with this lib, but with useState. Because with useAppForm hook I have problem to handle changes. And I also have this ts problem. I am not sure how to fix this. image
> > 
> > 
> > Could you push the latest changes to github? P.S. I didn't delete cursor helper, but I remember about it.
> 
> @v0ldemar01, I did it with no-verify flag because I have type errors P.S.I didn't delete the cursor helper, but I remember about it.

Got it, I will try to handle it soon

liza-veis:
let's use https://github.com/facebook/lexical instead of contenteditable lib

OlenaYatseniuk:
> let's use https://github.com/facebook/lexical instead of contenteditable lib

Okay, I will try to handle it tomorrow. 
Also, I will need to make another branch and reopen this pr, because I can't manage to resolve this conflicts.

v0ldemar01:
> > let's use https://github.com/facebook/lexical instead of contenteditable lib
> 
> Okay, I will try to handle it tomorrow. Also, I will need to make another branch and reopen this pr, because I can't manage to resolve this conflicts.

Please wait until tomorrow, maybe I found the core of current issue

v0ldemar01:
So @OlenaYatseniuk I fixed the issue with the initial cursor, everything works properly but not for lineToInsert(Enter key). We need to add a workaround for it. The main idea is checking tags in innerHTML and replace it with appropriate ones.

OlenaYatseniuk:
@v0ldemar01 Hey, thanks for your help! Actually, there is another problem =(
 When you put debounce outside of useEffect. I see that now, handleSaveNote is called every 2 seconds after you type something even when you don't type anything else.
I hope I and @HIMDanny will figure out how to fix it too.

v0ldemar01:
> @v0ldemar01 Hey, thanks for your help! Actually, there is another problem =( When you put debounce outside of useEffect. I see that now, handleSaveNote is called every 2 seconds after you type something even when you don't type anything else. I hope I and @HIMDanny will figure out how to fix it too.

Missed checker for isDirty form, it is easy to fix

OlesiaTr (dismissed):

HIMDanny:
@v0ldemar01, I have fixed the issue with endless server requests by declaring the `handleSaveNoteWithDebounce` in the useEffect. Before that, the function was recreated on the component's rerender, and as it was in the `useEffect` dependency of course it made an endless loop. Please tell me, if this is ok for you.

image

v0ldemar01:
> @v0ldemar01, I have fixed the issue with endless server requests by declaring the `handleSaveNoteWithDebounce` in the useEffect. Before that, the function was recreated on the component's rerender, and as it was in the `useEffect` dependency of course it made an endless loop. Please tell me, if this is ok for you.
> 
> image

As for me it's strange behavour of declaring debounce function inside useEffect. But we can proceed with it, it's not critical, the most wanted is lineBreak for now there

HIMDanny:
another way, I guess, is to use a debounce hook and change only values after a certain time. But now I will switch to the line break problem

shared/src/packages/journal/libs/validation-schemas/create-journal-entry.validation-schema.ts
```diff
@@ -0,0 +1,24 @@
...
+} from '../enums/enums.js';
+import { type JournalEntryCreateRequestDto } from '../types/types.js';
+
+const createJournalEntry = joi.object<JournalEntryCreateRequestDto, true>({
+  title: joi
+    .string()
+    .trim()
+    .required()
+    .max(JourneyEntryValidationRule.MAXIMUM_TITLE_LENGTH)
+    .min(JourneyEntryValidationRule.MINIMUM_TITLE_LENGTH)
```
dmytro-kucherenko: Is there a reason to set min `1` if you already have `required`? You even didn't set message for it because there's no way it will happen)

OlenaYatseniuk: You are right.

backend/src/packages/journal-entries/journal-entry.repository.ts
```diff
@@ -14,20 +14,38 @@ class JournalEntryRepository implements Repository {
...
 
-  public async findAll(): Promise<JournalEntryEntity[]> {
+  public async findAllByUserId(userId: number): Promise<JournalEntryEntity[]> {
     const journalEntries = await this.journalEntryModel
       .query()
-      .select()
-      .castTo<JournalEntryCommonQueryResponse[]>()
-      .execute();
+      .where({ userId })
+      .orderBy('updatedAt', 'DESC')
```
fshabanov: Create an enum for sort type

frontend/src/libs/enums/app-route.enum.ts
```diff
@@ -7,6 +7,7 @@ const AppRoute = {
   CHATS: '/chats',
   CHATS_$ID: '/chats/:id',
   JOURNAL: '/journal',
+  JOURNAL_ENTRY_$ID: '/journal/:id',
```
fshabanov:
```suggestion
  JOURNAL_$ID: '/journal/:id',
```

frontend/src/pages/journal/components/journal-sidebar/journal-sidebar.tsx
```diff
@@ -33,6 +39,15 @@ const JournalSidebar: React.FC<Properties> = ({
     },
   );
 
+  const handlePlusButtonClick = useCallback(() => {
```
fshabanov:
```suggestion
  const handleCreateJournalEntry = useCallback(() => {
```

fshabanov (changes requested):

frontend/src/pages/journal/components/note/note.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  const { field: titleField } = useFormController({ name: 'title', control });
+  const { field: textField } = useFormController({ name: 'text', control });
+  const { value: titleValue, onChange: onTitleChange } = titleField;
+  const { value: textValue, onChange: onTextChange } = textField;
+
+  const dispatch = useAppDispatch();
+  const { id } = useParams();
+
+  const debouncedTitleValue = useDebounce(titleValue, SAVE_NOTE_TIMEOUT);
```
Ekit18, line 63: Is it possible to use here already existing debounce function, which located in shared folder?

OlenaYatseniuk: No, you can see @HIMDanny`s conversation in Slack about it. We cannot use it with useEffect, but we need to have one reference to it.

Ekit18 (dismissed):

frontend/src/pages/journal/components/note/note.tsx
```diff
@@ -0,0 +1,179 @@
...
+      </div>
+      <div className={styles['text-area']}>
+        <LexicalComposer
+          initialConfig={{
+            namespace: 'NoteTextEditor',
+            onError: handleNoteError,
+            theme: {
+              paragraph: styles['paragraph'] as string,
+            },
+            editorState: textValue,
```
v0ldemar01: Without convertTextToNode wrapper

frontend/src/pages/journal/libs/constants/constants.ts
```diff
@@ -0,0 +1,8 @@
+const DEFAULT_NOTE_PAYLOAD = {
+  title: 'New note',
+  text: '{"root":{"children":[{"children":[{"detail":0,"format":0,"mode":"normal","style":"","text":"","type":"text","version":1}],"direction":"ltr","format":"","indent":0,"type":"paragraph","version":1}],"direction":"ltr","format":"","indent":0,"type":"root","version":1}}',
```
v0ldemar01: What is a purpose of it's text's default value

v0ldemar01 (changes requested):

frontend/src/pages/journal/components/note/note.tsx
HIMDanny: got it. I thought there would be comments if it wouldn't be wrapped, but happened the opposite)

frontend/src/pages/journal/libs/constants/constants.ts
HIMDanny: the lexical editor state needs to accept this structure for the initial value. If we would pass just text there would be an error. You can say, but we pass the title as the normal text. Yes, but we also send it as a simple text to the backend. By setting the text with simple text we would need to make a conversion like with title so that lexical would accept this data on the initial render. Then we would need to check whether this is a simple text from default values or actually a lexical structure from the backend which I think is an unnecessary operation. Please tell me if you have more questions)

frontend/src/pages/journal/components/note/note.tsx
HIMDanny: should I also unwrap `getTextContentFromEditorState` then?

frontend/src/pages/journal/libs/constants/constants.ts
v0ldemar01: I think the backend shouldn't know anything about our approach, btw it's strange to have different methods of conversioning. It had better save simple string, without json structure related to that lib

HIMDanny: first, the backend accepts it just as simple string text, I don't think this is a problem. But the main thing is that with converting to the simple text we will make the whole library unnecessary and remove all its features as new lines and so on.  Furthermore, I am pretty sure that we would need this structure so that lexical would make a rich text. I am talking about the rich text because we have an updated design for the journal which implies the use of it

v0ldemar01: Are you sure that mobile side will handle it as expected?

HIMDanny: about title. We need to convert to simple text because we have a restriction for title length which does not let to pass lexical structure. That is why we needed to convert it to simple text. And of course, I don't think that title should have some rich text features

v0ldemar01: I am just wondering about different approaches for title and text, it is confusing

HIMDanny: if there would the problems with mobile, we can convert it to HTML then, and send it to the backend

v0ldemar01: Yep, let do it)

HIMDanny: > I am just wondering about different approaches for title and text, it is confusing

I see. But let's assume we need to make the words bold, italic, etc. We wouldn't have this opportunity with simple text, but with lexical structure it is possible

HIMDanny: > Yep, let do it)

ok, then I will handle it

v0ldemar01: Good point, but it's overhead for this time)

frontend/src/pages/journal/components/note/components/note-input/note-input.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const handleChange = useCallback(
+    (newValue: string) => {
+      const parser = new DOMParser();
+      const dom = parser.parseFromString(newValue, 'text/html');
+
+      onChange(dom.body.textContent);
+    },
+    [onChange],
+  );
```
HIMDanny: it will also convert the note body into a simple text. It removes the whole sense of the editor then

frontend/src/pages/journal/components/note/components/note-input/libs/plugins/on-change/on-change.plugin.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import { NOTE_SANITIZER_OPTIONS } from '../../constants/constants.js';
+import { $generateHtmlFromNodes } from '../../helpers/helpers.js';
+import { useLexicalComposerContext } from '../../hooks/hooks.js';
+
+type Properties = {
+  onChange: (newValue: string) => void;
+};
+
+const OnChangePlugin: React.FC<Properties> = ({ onChange }) => {
```
liza-veis, line 12: Can't it be rewritten as a custom hook?

frontend/src/pages/journal/components/note/components/note-input/libs/helpers/helpers.ts
```diff
@@ -0,0 +1,7 @@
+export { $generateHtmlFromNodes, $generateNodesFromDOM } from '@lexical/html';
+export {
+  $createParagraphNode,
+  $createTextNode,
+  $getRoot,
+  $insertNodes,
```
liza-veis: let's rename all this helpers without $

frontend/src/pages/journal/components/note/components/note-input/note-input.tsx
```diff
@@ -0,0 +1,86 @@
...
+}: Properties<T>): JSX.Element => {
+  const { field } = useFormController({ name, control });
+  const { value, onChange } = field;
+
+  const handleChange = useCallback(
+    (newValue: string) => {
+      const parser = new DOMParser();
+      const dom = parser.parseFromString(newValue, 'text/html');
+
+      onChange(dom.body.textContent);
```
liza-veis: can't we just pass value instead of this?
```suggestion
      onChange(newValue);
```

frontend/src/pages/journal/components/note/components/note-input/note-input.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  return (
+    <div className={styles['text-area']}>
+      <LexicalComposer
+        initialConfig={{
+          namespace: `note-${name}`,
+          onError,
+          theme: { paragraph: styles['paragraph'] as string },
+          editorState: (editor: LexicalEditor): void => {
+            editor.update(() => {
```
liza-veis: let's move this outside jsx

liza-veis (changes requested):

frontend/src/pages/journal/components/note/components/note-input/note-input.tsx
v0ldemar01: Line break is working now) We need to merge this PR ASAP, so current scope doesn't contain advanced editor

frontend/src/pages/journal/components/note/components/note-input/libs/plugins/on-change/on-change.plugin.tsx
HIMDanny: to extend or change the logic in the Lexical we need to use such plugins. So I don't think that custom hook is the option

frontend/src/pages/journal/components/note/components/note-input/note-input.tsx
OlenaYatseniuk: Yep, I agree with @HIMDanny . Now logic with the new line doesn't work as expected. As it converts to simple text, it doesn't know anything about indentation after page reloading.

OlenaYatseniuk: @v0ldemar01 Hmmm, as I can see in my localhost, it works when you are typing, but after reloading you can see one line. Do you have another behavior?

frontend/src/pages/journal/components/note/components/note-input/libs/helpers/helpers.ts
HIMDanny: these are helper functions that show their intent as to where they can or cannot be used. So this prefix specifies that the function can only be used inside `.update` or `.read` methods. But if you think that it is necessary to rename them, I will do this

frontend/src/pages/journal/components/note/components/note-input/note-input.tsx
HIMDanny: @v0ldemar01, asked me to do the opposite here https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/293#discussion_r1332483149 )

frontend/src/pages/journal/components/note/components/note-input/note-input.tsx
HIMDanny: we can, and then the response will be presented in HTML string

frontend/src/pages/journal/components/note/components/note-input/libs/plugins/on-change/on-change.plugin.tsx
v0ldemar01: It's not possible, bcos it should be in composer context

frontend/src/pages/journal/components/note/components/note-input/note-input.tsx
HIMDanny: I don't think it works properly now)

How I typed:
image

After reload:
image

OlenaYatseniuk: Yeah, I told above about this.

v0ldemar01: Got it)

HIMDanny (dismissed):

HIMDanny (approved):

liza-veis (approved):

Issue #285, opened 2023-09-13
Remove unnessesary options from the Meditation screen


oksana-mlynska:
According to updated requirements 'Daily inspirations' and 'Favorites' functionality will be removed from Meditation page. So we only need to leave three options: Meditation, Relaxing sounds and Breath exercise.

![Image](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/140756353/7f603857-3617-4f98-97d4-f3f281d6f0d6)

Issue #282, opened 2023-09-13
Load meditation entries


liza-veis:
- [x] when user clicks on meditation item on meditation sidebar, all meditation entries should be loaded

k-mlch:
@iamAlinaaa 

1. there should be 3 topics (Meditation, Relaxing sounds and Breath exercise.)
2. and meditations should be loaded inside the Meditation topic

PR #297, opened 2023-09-15
cp-285: Remove Unnecessary Meditation Data

iamAlinaaa:
made all due to task #285  : According to updated requirements 'Daily inspirations' and 'Favorites' functionality will be removed from Meditation page. So we only need to leave three options: Meditation, Relaxing sounds and Breath exercise.

Снимок экрана 2023-09-15 124112


dmytro-kucherenko (approved):

danilittaa (approved):

mobile/src/screens/meditation-home/libs/constants.ts
```diff
@@ -1,13 +1,6 @@
 const mockedData = [
```
liza-veis, line 1: let's remove this, they shouldn't be mocked, they are different pages. And we only need to leave Meditation option for now

liza-veis (changes requested):

mobile/src/screens/meditation-home/libs/constants.ts
liza-veis: let's return this meditation screen by route meditation/list

OlesiaTr (approved):

mobile/src/screens/meditation-home/libs/constants.ts
iamAlinaaa: @Liza-Veis i guess i can done it better in this [task ](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/issues/282)related to meditation entries, because for now we can only mock the data and this functionality is built defferently. Later i will connect each page with id of meditation entry, so the mechanism will change. So should i do it in the task linked above ?

iamAlinaaa: as i asked on the meeting , we decided to make necessaty changes in next task about journal entries

PR #308, opened 2023-09-17
cp-306: Add Drop Shadows

danilittaa:
- added all shadows: 
image
image
image
image


mobile/src/screens/notification-settings/components/setting/styles.ts
iamAlinaaa, line 1: i guess the background should be white then or linear gradient, because background color that we are using for setting component doesnt distinguish our component, thats why shadow was added

OlesiaTr (dismissed):

mobile/src/screens/notification-settings/components/setting/styles.ts
danilittaa: After a quick chat with @k-mlch we decided to make the background on this page white (as in Figma) 
image

liza-veis (approved):

PR #310, opened 2023-09-17
cp-299: Mobile Sign Up Imporvements

iamAlinaaa:
1. Added fixed dark color for every phone model input
2. Reduced title text size for better layout
3. Added spacing for sign up screen at the bottom
Снимок экрана 2023-09-17 190919
Снимок экрана 2023-09-17 190910

### DIFF
--- a/mobile/src/screens/auth/components/sign-up-form/styles.ts
+++ b/mobile/src/screens/auth/components/sign-up-form/styles.ts
@@ -9,7 +9,7 @@ const styles = StyleSheet.create({
     paddingHorizontal: 55,
   },
   titleText: {
-    fontSize: 35,
+    fontSize: 30,
     color: AppColor.WHITE,
     marginBottom: 30,
     marginTop: 30,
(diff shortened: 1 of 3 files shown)


mobile/src/screens/auth/components/sign-up-form/styles.ts
```diff
@@ -7,9 +7,10 @@ const styles = StyleSheet.create({
     height: '100%',
     width: '100%',
     paddingHorizontal: 55,
+    marginBottom: 55,
```
danilittaa, line 10: Isn't it better to center all the content? (We don't know the user's screen resolution, and the layout may look different on various devices, so I suggest centering everything)

iamAlinaaa: @danilittaa if i do it so (center content), it will be just looks like this, because i was makng some improvements for sign background task with puprose to make scrollin bar at the edge< and i just moved this fixed padding from sign background. input field shouldnt have fixed width, so its look okay with screen padding
Снимок экрана 2023-09-18 113346

liza-veis (approved):

mobile/src/screens/auth/components/sign-up-form/styles.ts
liza-veis: it's better not to center screens where content may exceed screen height

PR #311, opened 2023-09-18
cp-300: Fix sidebar redirect

HIMDanny:
- Added query string enums
- Added a helper to add a query string to the URL
- Updated `useSidebarShow` hook
- Removed query strings in chat component links


frontend/src/libs/components/navigation-menu-wrapper/libs/constants.ts
```diff
@@ -1,12 +1,25 @@
...
+});
+
 const SIDEBAR_ROUTES: Route[] = [
   { path: AppRoute.ROOT, name: 'home', icon: 'home' },
-  { path: AppRoute.MEDITATION, name: 'meditation', icon: 'meditation' },
-  { path: AppRoute.CHATS, name: 'chats', icon: 'chats' },
-  { path: AppRoute.JOURNAL, name: 'journal', icon: 'journal' },
-  { path: AppRoute.PROFILE, name: 'profile', icon: 'user' },
+  {
+    path: urlWithQueryString(AppRoute.MEDITATION),
```
v0ldemar01: Can't we handle it via callback of setSidebarMode?

v0ldemar01 (changes requested):

frontend/src/libs/enums/query-strings/sidebar-mode-query-string-value.enum.ts
```diff
@@ -0,0 +1,6 @@
+const SidebarModeQueryStringValue = {
```
liza-veis:
```suggestion
const SidebarMode = {
```

frontend/src/libs/helpers/get-url-with-query-string.helper.ts
```diff
@@ -0,0 +1,16 @@
+import { type AppQueryStringKey, type AppRoute } from '#libs/enums/enums.js';
+import { type ValueOf } from '#libs/types/types.js';
+
+type QueryParameters = Record<ValueOf<typeof AppQueryStringKey>, string>;
+
+const getUrlWithQueryString = (
```
liza-veis, line 6: should accept path and parameters and return url, not function

frontend/src/pages/chat/components/chat-sidebar/chat-sidebar.tsx
```diff
@@ -72,10 +70,10 @@ const ChatSidebar: React.FC<Properties> = ({
         </div>
         <div className={styles['chat-list']}>
           {filteredElements.map((filteredChat) => {
-            const chatLink = (AppRoute.CHATS_$ID.replace(
+            const chatLink = AppRoute.CHATS_$ID.replace(
               ':id',
               String(filteredChat.id),
-            ) + '?sidebarMode=hide') as typeof AppRoute.CHATS_$ID;
+            ) as typeof AppRoute.CHATS_$ID;
```
liza-veis:
```suggestion
            ) as ValueOf<typeof AppRoute>;
```

liza-veis (changes requested):

frontend/src/libs/helpers/get-url-with-query-string.helper.ts
HIMDanny: oh, I see. Ok, will rewrite it

frontend/src/libs/components/navigation-menu-wrapper/libs/constants.ts
HIMDanny: but setSidebarMode is used in useSidebarState, and we can use hooks only in components. I didn't get what you mean

HIMDanny: moreover, the logic of the hook is focused on button clicks. not on links

liza-veis (approved):

frontend/src/pages/chat/chat.tsx
```diff
@@ -1,16 +1,23 @@
...
 
 import { ChatLayout, ChatSidebar } from './components/components.js';
 import styles from './styles.module.scss';
 
 const Chat: React.FC = () => {
+  const navigate = useNavigate();
   const { isSidebarShown, setIsSidebarShown } = useSidebarState();
 
   const handleBackButtonPress = useCallback(() => {
+    navigate(AppRoute.CHATS);
```
v0ldemar01, line 18: But we have appActions for such purpose

v0ldemar01 (changes requested):

frontend/src/pages/chat/chat.tsx
HIMDanny: unfortunately, it doesn't work properly like this. It doesn't add the query string after redirection with `appActions`

HIMDanny: so the option is to use the navigation from react-router like this

v0ldemar01: But there is no query string in `navigate(AppRoute.CHATS);`

v0ldemar01: Then how does it work?

HIMDanny: yep, but there is `setIsSidebarShown(true)` below

HIMDanny: I guess navigate in appActions rewrites the URL after applying the query string

OlenaYatseniuk (approved):
Looks good!

frontend/src/pages/chat/chat.tsx
v0ldemar01: navigate in appActions has the same logic as yours navigate. Btw, I've checked locally and both solutions work equally

HIMDanny: hm, that's strange, cause I've also tested. Did you write like this?
```ts
  const handleBackButtonPress = useCallback(() => {
    dispatch(appActions.navigate(AppRoute.CHATS));
    setIsSidebarShown(true);
  }, [setIsSidebarShown, dispatch]);
```

v0ldemar01: Ok, let's proceed with your solution, as we have not enough time to re-check it

v0ldemar01 (approved):

fshabanov (approved):

PR #319, opened 2023-09-19
cp-159: Add more steps to the survey#159

BrianSammit:
-  Add more steps to the survey according to mockups
-  Add a back button under the continue button 
- The back button only shows when there is a previous step
- Request to save the survey should be sent at the end of it after the last step

I'm waiting for the back-end part https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/issues/156 to test the proper save in the database


OlesiaTr (dismissed):

mobile/src/screens/survey/components/journaling-step/journaling-step.tsx
```diff
@@ -0,0 +1,128 @@
...
+      handleSubmitSurvey(getSurveyCategories(payload));
+    },
+    [handleSubmitSurvey],
+  );
+
+  const handleFormSubmit = useCallback((): void => {
+    void handleSubmit(handleSurveySubmit)();
+  }, [handleSubmit, handleSurveySubmit]);
+  const handleBack = (): void => {
+    navigation.navigate(SurveyScreenName.MEDITATION_EXPERIENCE);
```
dmytro-kucherenko: Each step has lots of similarities with others, at least I can see only another navigation route is a difference. There's no reason to duplicate so much code, so I think you can create a wrapper, which will contain all this logic with parent route `/survey`, and then change only the questions according to the children route, like `/survey/preferences-step`

BrianSammit: I will check that the problem with the route approach is that React Native does not use a route as in React. React Native uses stacks. But you're right I will find a way to not use the same code in the steps.

frontend/src/pages/surveys/survey.tsx
```diff
@@ -29,6 +29,11 @@ const Survey: React.FC = () => {
       void dispatch(
         authActions.createUserSurvey({
           userId: userId,
+          journaling_experience: [],
```
v0ldemar01: Use camel case here

mobile/src/libs/types/navigation/survey-navigation-parameter-list.type.ts
```diff
@@ -0,0 +1,12 @@
+import { type SurveyScreenName } from '#libs/enums/enums';
+
+type SurveyNavigationParameterList = {
+  [SurveyScreenName.PREFERENCES]: undefined | object;
+  [SurveyScreenName.FEELINGS]: undefined | object;
+  [SurveyScreenName.GOALS]: undefined | object;
+  [SurveyScreenName.WORRIES]: undefined | object;
+  [SurveyScreenName.MEDITATION_EXPERIENCE]: undefined | object;
+  [SurveyScreenName.JOURNALING_EXPERIENCE]: undefined | object;
```
v0ldemar01, line 9: Provide more specific types

mobile/src/screens/survey/components/feelings-step/feelings-step.tsx
v0ldemar01, line 1: Create a generic component for reusing, it is not a good practise of your approach

mobile/src/screens/survey/survey.tsx
```diff
@@ -1,41 +1,91 @@
...
+
+const SurveyStack = createNativeStackNavigator<SurveyNavigationParameterList>();
 
 const Survey: React.FC = () => {
   const dispatch = useAppDispatch();
+  const [preferencesSurvey, setPreferencesSurvey] = useState([]);
+  const [feelingsSurvey, setFeelingsSurvey] = useState([]);
+  const [goalsSurvey, setGoalsSurvey] = useState([]);
+  const [worriesSurvey, setWorriesSurvey] = useState([]);
+  const [meditationSurvey, setMeditationSurvey] = useState([]);
```
v0ldemar01: useAppForm instead

mobile/src/screens/survey/survey.tsx
```diff
@@ -1,41 +1,91 @@
...
-  );
+  const handleSubmit = (option: string[]): void => {
+    void dispatch(
+      authActions.createUserSurvey({
+        userId: userId,
+        preferences: preferencesSurvey,
+        feelings: feelingsSurvey,
+        goals: goalsSurvey,
+        worries: worriesSurvey,
+        meditation_experience: meditationSurvey,
```
v0ldemar01: Same for camel case

v0ldemar01 (changes requested):
Also please contact to Oscar to be in touch with current changes

iamAlinaaa (commented):
We already have fonts in our mobile app, please add fonts everywhere where necessary

Issue #316, opened 2023-09-18
Reduce the size of the Chat name on the Chat screen


oksana-mlynska:
Maybe the size of the chat name on the Chat screen should be reduced, it looks much smaller on the design and, also, when it is too big, it is located too close to the 'Back' arrow

![Image](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/140756353/f08e52f2-6bd0-46c9-90aa-f8b9003a22cd)

PR #323, opened 2023-09-19
cp-316: Mobile Chat Name Size Reduction

iamAlinaaa:
1. Reduced the size of the chat name on the Chat screen to look like on design and have good layout for mobile.
2. Now it is not located too close to the 'Back' arrow
All changes due to #316 

Снимок экрана 2023-09-19 184232


mobile/src/libs/components/header/header.tsx
```diff
@@ -49,7 +51,9 @@ const Header: React.FC<Properties> = ({
       {isArrowVisible && <BackButton />}
 
       <View style={styles.titleBadgeContainer}>
-        <Text style={styles.title}>{title ?? name}</Text>
+        <Text style={isChat ? styles.chatTitle : styles.title}>
```
dmytro-kucherenko:
```suggestion
        <Text style={[styles.title, isChat && styles.chatTitle]}>
```
Styles have similarities

OlesiaTr (dismissed):

dmytro-kucherenko (dismissed):

mobile/src/libs/components/header/header.tsx
```diff
@@ -20,13 +20,15 @@ type Properties = {
   isArrowVisible?: boolean;
   badgeCount?: number;
   isProfileVisible?: boolean;
+  isChat?: boolean;
```
danilittaa: Why did you only change the sizes for the chat page?

iamAlinaaa: @danilittaa because my task was only about it

danilittaa: Perhaps the QA team simply doesn't realize that the same component is used everywhere. Why not make these changes in the header once so that we don't have to fix it on every page later?😅

iamAlinaaa: @danilittaa  i guess that change is need to be done only here 🤔 because its chat name (not page named chat with all chats) and this name can be different every time, it can be quite long etc. and if we speak about our headers for screens like meditation or notes , they have appropriate size and they look good. we only can have problems with chat header for now

liza-veis: we need to rename this property to something like - fontSize: 'small' | 'large'. Header shouldn't know about where it will be used

liza-veis (changes requested):

BrianSammit (dismissed):

OlesiaTr (dismissed):

OlenaYatseniuk (dismissed):

v0ldemar01 (dismissed):

fshabanov (dismissed):

mobile/src/libs/components/header/header.tsx
```diff
@@ -20,13 +20,15 @@ type Properties = {
...
   isProfileVisible?: boolean;
+  fontSize?: 'small' | 'large';
 };
 
 const Header: React.FC<Properties> = ({
   title,
   isArrowVisible = false,
   badgeCount = DEFAULT_BADGE_COUNT,
   isProfileVisible = false,
+  fontSize,
```
liza-veis: let's make it large by default and remove from components where it's used
```suggestion
  fontSize = 'large',
```

liza-veis (changes requested):

dmytro-kucherenko (approved):

v0ldemar01 (approved):

fshabanov (approved):

liza-veis (approved):

PR #325, opened 2023-09-19
cp-304: Card view with long titles. Fix

OlesiaTr:


frontend/src/libs/components/card/styles.module.scss
```diff
@@ -72,6 +71,12 @@
   word-break: break-word;
 }
 
+@media (width >= 900px) {
```
liza-veis: we use < sign everywhere, so here we should use it too

liza-veis (changes requested):

frontend/src/libs/components/card/styles.module.scss
OlesiaTr: Would it be ok if I put 899 then? Because without = it will be broken on 900px

liza-veis: you can use =, but we need to invert the sign <= will be ok

OlesiaTr: Ohhh, but can I use strict measures then?

OlesiaTr: I can write it like this

`@media (900px <= width) `

And it's working, but it feels so wrong😖. And I used to write it more like this `@media (min-width: 900px)`

liza-veis: I also like this option `@media (min-width: 900px)`, but we use this `@media (900px <= width)` in other places in the app

OlesiaTr: Ye, I was asking cause didn't see in the development this way of writing it yet. Ok, I changed it!

OlenaYatseniuk (dismissed):

liza-veis (approved):

v0ldemar01 (dismissed):

frontend/src/libs/components/card/styles.module.scss
```diff
@@ -73,6 +72,12 @@
   word-break: break-word;
 }
 
+@media (900px <= width) {
```
liza-veis: we need to join it with a query below for < 900px

liza-veis (changes requested):

OlesiaTr:
With changes looks like this

https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/104642736/c9f99e45-6617-4c57-8c11-fae3641be9a6

dmytro-kucherenko (approved):

OlenaYatseniuk (approved):

v0ldemar01 (approved):

fshabanov (approved):

liza-veis (approved):

PR #307, opened 2023-09-17
cp-295: Deleting note functionality

Ekit18:
https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/92929504/3aebf217-a2a8-4ff5-a91d-e497d7dd30bb
image
image

PR #327, opened 2023-09-19
cp-268: Delete chat functionality (reopened)

OlenaYatseniuk:
Reopened this pr due to merge conflicts in the previous one.


backend/src/packages/chats/chat.service.ts
```diff
@@ -83,8 +92,24 @@ class ChatService implements Service {
...
+    id,
+    userId,
+  }: {
+    id: number;
+    userId: number;
+  }): Promise<boolean> {
+    const allChats = await this.findAllByUserId(userId);
+
+    const chat = allChats.items.find((chat) => {
+      return chat.id === Number(id);
```
dmytro-kucherenko:
```suggestion
      return chat.id === id;
```
You already receive it as number

frontend/src/libs/components/card/card.tsx
```diff
@@ -20,37 +22,52 @@ const Card: React.FC<Properties> = ({
...
+        onClick={onClick}
+      >
+        <div className={styles['item-info']}>
+          {!hasNoImageOrIcon && (
+            <div className={styles['image-container']}>
+              {hasImage && (
+                <div className={styles['image-placeholder']}>
+                  <img
+                    src={imageUrl}
+                    alt="not found"
```
dmytro-kucherenko: This text isn't quite representative in case the client will see it

frontend/src/libs/components/card/card.tsx
```diff
@@ -20,37 +22,52 @@ const Card: React.FC<Properties> = ({
...
+              )}
             </div>
           )}
+          <div className={styles['title']}>{title}</div>
         </div>
+      </button>
+      {iconRight && (
+        <button className={styles['icon-right']} onClick={onIconClick}>
+          <Icon name={iconRight} color={iconColor} />
+        </button>
```
dmytro-kucherenko: Is it possible to use common `Button` component here?

frontend/src/libs/components/modal/modal.tsx
```diff
@@ -37,18 +38,21 @@ const Modal: React.ForwardRefRenderFunction<
...
-                iconColor={IconColor.BLACK}
-                iconWidth={30}
-                iconHeight={30}
-                style="rounded-transparent"
-                onClick={handleClose}
-                isLabelVisuallyHidden
-              />
-            </div>
+
+            {showCrossIcon && (
```
dmytro-kucherenko:
```suggestion
            {isCloseIconShown && (
```
Actually we renamed the icon

frontend/src/pages/chat/components/delete-chat-modal/delete-chat-modal.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  const { isLoading } = useAppSelector(({ chats }) => {
+    return {
+      isLoading: chats.deleteChatDataStatus === DataStatus.PENDING,
+    };
+  });
+
+  const handleDeleteChatEntry = useCallback(() => {
+    void dispatch(chatActions.deleteChat(id as number)).finally(() => {
+      chatDeleteModalReference.current?.close();
```
dmytro-kucherenko, line 43: It won't work, you have to close it on `FULFILLED` data status

frontend/src/slices/chats/actions.ts
```diff
@@ -59,4 +60,29 @@ const createMessage = createAsyncThunk<
...
+  const { chatApi } = extra;
+  await chatApi.deleteChat(id);
+
+  const {
+    chats: { chats },
+  } = getState();
+
+  dispatch(appActions.navigate(AppRoute.CHATS));
+
+  return chats.filter((chat) => {
```
dmytro-kucherenko: You can do it in the slice in case of `deleteChat.fulfilled`. At least it's approach we already get used to and also you wouldn't use exta stuff, like `getState`

frontend/src/pages/chat/components/delete-chat-modal/delete-chat-modal.tsx
HIMDanny: If I am not mistaken, to make it work we should add `.unwrap()` after `dispatch()`. But if there is a more clear solution of course we should try it first

backend/src/packages/chats/chat.controller.ts
```diff
@@ -302,6 +331,55 @@ class ChatController extends BaseController {
...
     };
   }
+
+  /**
+   * @swagger
+   * /chats/{id}:
+   *   delete:
+   *     description: Delete chat by id
+   *     security:
+   *      - bearerAuth: []
```
HIMDanny, line 341: Let's also add path parameter
```suggestion
   * /chats/{id}:
   *   delete:
   *     description: Delete chat by id
   *     security:
   *      - bearerAuth: []
   *     parameters:
   *       -  in: path
   *          description: Chat id
   *          name: id
   *          required: true
   *          type: number
   *          minimum: 1
```

backend/src/packages/chats/chat.repository.ts
```diff
@@ -74,8 +93,8 @@ class ChatRepository implements Repository {
     return Promise.resolve(null);
   }
 
-  public delete(): ReturnType<Repository['delete']> {
-    return Promise.resolve(false);
+  public async delete(id: number): Promise<number> {
+    return await this.chatModel.query().deleteById(id).execute();
   }
```
HIMDanny:
```suggestion
  public delete(id: number): Promise<number> {
    return this.chatModel.query().deleteById(id);
  }
```

backend/src/packages/chats/chat.service.ts
```diff
@@ -32,6 +33,14 @@ class ChatService implements Service {
     return Promise.resolve(null);
   }
 
+  public async findById(
+    id: number,
+  ): Promise<ReturnType<ChatEntity['toObject']> | null> {
+    const chat = await this.chatRepository.findById(id);
+
+    return chat?.toObject() ?? null;
+  }
```
HIMDanny: do we actually need this?

OlesiaTr (dismissed):

frontend/src/pages/chat/components/delete-chat-modal/delete-chat-modal.tsx
dmytro-kucherenko: I'm not sure how `unwrap` will work in case of rejection, but basically it's also a good option

backend/src/libs/types/service.type.ts
```diff
@@ -5,7 +5,7 @@ type Service<T = unknown> = {
   }>;
   create(payload: unknown): Promise<T>;
   update(): Promise<T>;
-  delete(): Promise<boolean>;
+  delete(payload: unknown): Promise<boolean>;
```
fshabanov:
```suggestion
  delete(id: number): Promise<boolean>;
```

backend/src/libs/types/repository.type.ts
```diff
@@ -3,7 +3,7 @@ type Repository<T = unknown> = {
   findAll(): Promise<T[]>;
   create(payload: unknown): Promise<T>;
   update(payload: unknown): Promise<T>;
-  delete(): Promise<boolean>;
+  delete(id: number): Promise<number>;
```
fshabanov: Why do we return number here?

fshabanov (changes requested):

backend/src/packages/chats/chat.service.ts
OlenaYatseniuk: Actually, it is a string as it is a param. But I'll move this logic to the controller. Thanks for noticing.

frontend/src/libs/components/card/card.tsx
OlenaYatseniuk: Actually, it has already existed in development, but it is not a problem to rewrite the message.

frontend/src/pages/chat/components/delete-chat-modal/delete-chat-modal.tsx
OlenaYatseniuk: But if we have rejected status, I think we also need close it.

frontend/src/slices/chats/actions.ts
OlenaYatseniuk: I am not sure that it is a good way to do it in the slice. Slice must be as simple as it can be. Also, I need to pass deletedId to slice to make this logic there, so it looks weird to me.

backend/src/libs/types/service.type.ts
OlenaYatseniuk: But I need to pass userId too. Isn't it better to have 1 payload instead of 2 params : id and userId?

backend/src/libs/types/repository.type.ts
OlenaYatseniuk: `  public delete(id: number): Promise<number> {
    return this.chatModel.query().deleteById(id).execute();
  }`
  Because this will return a number, so that's why in service: I have this `return Boolean(await this.chatRepository.delete(id));`

backend/src/packages/chats/chat.service.ts
```diff
@@ -83,8 +83,24 @@ class ChatService implements Service {
...
-    return Promise.resolve(true);
+  public async delete({
+    id,
+    userId,
+  }: {
+    id: number;
+    userId: number;
+  }): Promise<boolean> {
+    const allChats = await this.findAllByUserId(userId);
+
```
Ekit18: maybe it is better to do smth like this:
    const chat = await this.find(id); 
    if (chat.userId !== userId) {
      throw new ChatError({
        status: HTTPCode.BAD_REQUEST,
        message: ExceptionMessage.INCORRECT_CREDENTIALS,
      });
    }
    return Boolean(await this.chatRepository.delete(id));

backend/src/packages/chats/chat.service.ts
Ekit18: do you mean findById func? If yes we need this because without it we can`t check who is trying to delete this chat

Ekit18 (dismissed):

backend/src/packages/chats/chat.service.ts
```diff
@@ -83,8 +96,26 @@ class ChatService implements Service {
...
+  public async delete({
+    id,
+    userId,
+  }: {
+    id: number;
+    userId: number;
+  }): Promise<boolean> {
+    const chat = await this.findById(id);
+    if (!chat) {
+      return false;
```
v0ldemar01: Throw the not found error

backend/src/packages/chats/chat.service.ts
```diff
@@ -83,8 +96,26 @@ class ChatService implements Service {
...
+    }
+
+    if (chat.members[FIRST_ARRAY_INDEX]?.userId !== userId) {
+      throw new ChatError({
+        status: HTTPCode.BAD_REQUEST,
+        message: ExceptionMessage.INCORRECT_CREDENTIALS,
+      });
+    }
```
v0ldemar01: As we are not having permissions, let's proceed with 404 status code

frontend/src/libs/components/card/card.tsx
```diff
@@ -20,37 +22,55 @@ const Card: React.FC<Properties> = ({
...
+                <div className={styles['icon-background']}>
+                  <Icon name={iconName as IconName} color={iconColor} />
+                </div>
+              )}
             </div>
           )}
+          <div className={styles['title']}>{title}</div>
         </div>
+      </button>
+      {iconRight && (
```
v0ldemar01: Add boolean for it

frontend/src/pages/chat/components/chat-sidebar/chat-sidebar.tsx
```diff
@@ -33,12 +36,19 @@ const ChatSidebar: React.FC<Properties> = ({
 }) => {
   const { id } = useParams<{ id: string }>();
   const dispatch = useAppDispatch();
+  const [deletedChat, setDeleteChat] = useState<null | number>(null);
```
v0ldemar01:
```suggestion
  const [chatToDelete, setChatToDelete] = useState<null | number>(null);
```

v0ldemar01 (changes requested):

frontend/src/libs/components/card/card.tsx
OlenaYatseniuk: @v0ldemar01 But if I make it a boolean, I need to pass IconName to have the name of the icon, so in this way, the code under will be displayed, but I don't need this icon there, so as a result I see this big icons between picture and title, but I don't want it. So I think my previous solution looks ok. What do you think?
image
image

frontend/src/pages/chat/components/delete-chat-modal/delete-chat-modal.tsx
dmytro-kucherenko: I think in case of getting rejected status client should see error toast, but modal still have to be opened

frontend/src/slices/chats/actions.ts
dmytro-kucherenko: Actually I would pass the id in similar situation, but it's quite controversial question

backend/src/libs/types/service.type.ts
Dmytro-Kucherenko: I think you can pass `{ id: number, userId: number }`then

frontend/src/slices/chats/actions.ts
OlenaYatseniuk: So maybe @v0ldemar01 or @fshabanov can clarify for me, where would it be better to place this logic: in action or slice?

frontend/src/pages/chat/components/delete-chat-modal/delete-chat-modal.tsx
OlenaYatseniuk: Ok, I can do it

frontend/src/libs/components/card/card.tsx
v0ldemar01: const hasIconRight = Boolean(iconRight) of something like that or actionIcon looks better

OlenaYatseniuk: Now I see your point, agree.

backend/src/libs/types/repository.type.ts
```diff
@@ -3,7 +3,7 @@ type Repository<T = unknown> = {
   findAll(): Promise<T[]>;
   create(payload: unknown): Promise<T>;
   update(payload: unknown): Promise<T>;
-  delete(): Promise<boolean>;
+  delete(id: number): Promise<number | boolean>;
```
v0ldemar01: It should be only one type for returning

frontend/src/slices/chats/actions.ts
v0ldemar01: It is ok to stay with current approach

backend/src/packages/chat-messages/chat-message.repository.ts
```diff
@@ -60,7 +60,10 @@ class ChatMessageRepository implements Repository {
   }
 
   public delete(): ReturnType<Repository['delete']> {
-    return Promise.resolve(true);
+    //TODO
+    const deletedId = 0;
```
v0ldemar01: Is it mocked?

v0ldemar01 (changes requested):

frontend/src/slices/chats/actions.ts
OlenaYatseniuk: Ok, thanks!

backend/src/packages/chat-messages/chat-message.repository.ts
OlenaYatseniuk: Oh, i just missed this repository to change

backend/src/libs/types/repository.type.ts
OlenaYatseniuk: @v0ldemar01 So, should we return back (as it was in the beginning) in our delete repository to this to return boolean everywhere?
image

OlenaYatseniuk: So I changed it to have boolean everywhere

v0ldemar01: Let's delete in repository level will return number)

OlenaYatseniuk: Ok

backend/src/packages/chat-messages/chat-message.repository.ts
OlenaYatseniuk: So I returned it back to  return number in repository, so it is needed everywhere now. You can see this also here #307

v0ldemar01 (dismissed):

backend/src/packages/chat-messages/chat-message.repository.ts
```diff
@@ -60,7 +60,9 @@ class ChatMessageRepository implements Repository {
   }
 
   public delete(): ReturnType<Repository['delete']> {
-    return Promise.resolve(true);
+    const deletedId = 0;
```
liza-veis: isn't it a deletedCount?

frontend/src/packages/chats/chats-api.ts
```diff
@@ -45,6 +45,20 @@ class ChatApi extends BaseHttpApi {
...
     return await response.json<ChatGetAllItemResponseDto>();
   }
+
+  public async deleteChat(id: number): Promise<boolean> {
+    const response = await this.load(
+      this.getFullEndpoint(ChatsApiPath.$ID, { id: id.toString() }),
+      {
+        method: 'DELETE',
+        contentType: ContentType.JSON,
+        payload: JSON.stringify({}),
```
liza-veis:
```suggestion
```

frontend/src/pages/chat/components/chat-sidebar/styles.module.scss
```diff
@@ -1,3 +1,7 @@
+.wrapper {
+  height: 83px;
+}
```
liza-veis:
```suggestion
```

frontend/src/slices/chats/actions.ts
```diff
@@ -59,4 +60,29 @@ const createMessage = createAsyncThunk<
...
+  ChatGetAllItemResponseDto[],
+  number,
+  AsyncThunkConfig
+>(`${sliceName}/delete-chat`, async (id, { extra, getState, dispatch }) => {
+  const { chatApi } = extra;
+  await chatApi.deleteChat(id);
+
+  const {
+    chats: { chats },
+  } = getState();
```
liza-veis: rewrite it to format that we use in other places

frontend/src/slices/chats/actions.ts
liza-veis: this should be done inside reducer

liza-veis (changes requested):

frontend/src/slices/chats/actions.ts
OlenaYatseniuk: @Liza-Veis Sorry, but I don't understand what you mean. Here https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/327#discussion_r1335150470 @v0ldemar01 agreed that it is okay to use getState and filter here, but not in slice.

frontend/src/slices/chats/actions.ts
OlenaYatseniuk: Could you, please, check this comment by @v0ldemar01 https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/327#discussion_r1335150470

v0ldemar01: @OlenaYatseniuk please follow @Liza-Veis comment, I have mistaken with it

frontend/src/slices/chats/actions.ts
liza-veis: i mean to use it like this :
```
 const { chats } = getState(({ chats }) => ({
   chats: chats.chats
 }));
```

frontend/src/slices/chats/actions.ts
OlenaYatseniuk: Ok, I'll move it to slice

v0ldemar01 (dismissed):

backend/src/packages/chats/chat.service.ts
```diff
@@ -83,8 +96,31 @@ class ChatService implements Service {
...
+  }): Promise<boolean> {
+    const chat = await this.findById(id);
+    if (!chat) {
+      throw new ChatError({
+        status: HTTPCode.NOT_FOUND,
+        message: ExceptionMessage.CHAT_NOT_FOUND,
+      });
+    }
+
+    if (chat.members[FIRST_ARRAY_INDEX]?.userId !== userId) {
```
liza-veis: what if user is not in first index?

liza-veis (changes requested):

backend/src/packages/chats/chat.service.ts
liza-veis: 1. user can have any position in members array 
2. let's do this in query (query only chats where user is included in)

OlenaYatseniuk: It is not about it. In redux we have this structure. So findById method will return this structure. And we will have this if we have a chat and we will have it, because the error will be thrown earlier.
image

liza-veis: it's a backend package, not frontend package and as I see here you check if chat related to user by checking if he is the first member, but we need to check if he is included in members

liza-veis: we don't even need findById method, we should delete by query and check if the deleted count is 1. Take a look at this pr where userId check also will be added in query https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/387

OlenaYatseniuk: Yes. You can see the response when I try to delete it from frontend. So I always have this first member with userId
image

OlenaYatseniuk: Oh, some of your messages are shown only now. So, I will think about it.

OlesiaTr (dismissed):

v0ldemar01 (approved):

liza-veis (approved):

PR #340, opened 2023-09-21
cp-269: Mobile Delete Chat Functionality

iamAlinaaa:
1. Done card swipe functionality which shows modal where user can change remove element or not.
2. Done chat delete functionality

https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/111873795/377dea95-28cc-49be-be00-be904666dc16


mobile/src/libs/components/button/styles.ts
```diff
@@ -53,6 +59,19 @@ const styles = StyleSheet.create({
...
+    marginHorizontal: 10,
+    paddingHorizontal: 20,
+  },
+
+  buttonModalDelete: {
+    backgroundColor: AppColor.BLUE_300,
+    borderRadius: 8,
+    marginHorizontal: 10,
+    paddingHorizontal: 20,
+  },
```
fshabanov: We can create a common style for both and then add backgroundColor based on the button types

mobile/src/libs/components/card/card.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import { type ImageSourcePropType } from 'react-native';
+import { Swipeable } from 'react-native-gesture-handler';
```
fshabanov: Import from #libs folder

mobile/src/libs/components/modal/modal.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+const Modal: React.FC<Properties> = ({
+  children,
+  isVisible,
+  onClose,
+  onDelete,
+  type,
+}) => {
+  return (
+    <RNModal visible={isVisible} transparent={true} animationType="none">
```
fshabanov:
```suggestion
    <RNModal visible={isVisible} transparent animationType="none">
```

mobile/src/libs/components/modal/modal.tsx
```diff
@@ -0,0 +1,60 @@
...
+                      label="Delete"
+                      onPress={onDelete}
+                      type="modal-delete"
+                    />
+                  </View>
+                </>
+              ) : (
+                <>
+                  <Pressable onPress={onClose} style={styles.closeButton}>
+                    {/* add icon here */}
```
fshabanov: Fix this comment

fshabanov (changes requested):

OlesiaTr (dismissed):

mobile/src/libs/components/modal/modal.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { Modal as RNModal } from 'react-native';
+
+import { Button, Pressable, Text, View } from '#libs/components/components';
+
+import { styles } from './styles';
+
+type Properties = {
+  isVisible: boolean;
+  onClose: () => void;
+  onDelete?: () => void;
```
dmytro-kucherenko, line 18: Maybe it's better to name it `onSubmit`? All in all it's common component, in local case submiting will be the deletion

mobile/src/screens/chat-list/chat-list.tsx
```diff
@@ -42,6 +46,20 @@ const ChatList: React.FC = () => {
...
+    setChatId(id);
+    setIsModalVisible(true);
+  };
+
+  const hanleCloseModal = (): void => {
+    setIsModalVisible(false);
+  };
+
+  const handleDeleteChat = (): void => {
+    setIsModalVisible(false);
```
dmytro-kucherenko:
```suggestion
    hanleCloseModal();
```

mobile/src/libs/components/modal/modal.tsx
iamAlinaaa: @Dmytro-Kucherenko this component is used for adding meditation entries, so i guess its better leave onDelete

mobile/src/libs/components/card/card.tsx
v0ldemar01: Not resolved

v0ldemar01 (changes requested):
Please resolve last comment

v0ldemar01 (dismissed):

mobile/src/libs/components/card/card.tsx
```diff
@@ -6,52 +6,65 @@ import {
...
   onPress,
+  iconRight,
+  onIconPress,
 }) => {
-  const handlePress = (): void => {
-    onPress(title, id);
+  const renderRightSwipeActions = (): React.ReactNode => {
+    return (
+      Boolean(iconRight) &&
+      iconRight && (
```
fshabanov:
```suggestion
      Boolean(iconRight) && (
```

mobile/src/screens/chat-list/chat-list.tsx
```diff
@@ -35,6 +37,8 @@ const ChatList: React.FC = () => {
   const dispatch = useAppDispatch();
   const navigation =
     useNavigation<NativeStackNavigationProp<ChatNavigationParameterList>>();
+  const [isModalVisible, setIsModalVisible] = useState<boolean>(false);
+  const [chatId, setChatId] = useState<number>();
```
fshabanov: You specify it is a number but don't give a default value, so it can be `undefined` as well 🤔

fshabanov (changes requested):

mobile/src/libs/components/card/card.tsx
iamAlinaaa: @fshabanov i had this issue in cp-296, and  Boolean(iconRight) && is not work properly i had an error, so i wrote it like this

fshabanov: @iamAlinaaa  What was the error? It doesn't really make sense to write `Boolean(iconRight) && iconRight` because you are checking one condition twice

mobile/src/screens/chat-list/chat-list.tsx
iamAlinaaa: @fshabanov when dispatching this chatId i dispatch this as number  void dispatch(chatsActions.deleteChat(chatId as number)); and i guess adding extra number to useState is not that useful, as we just can add any not necessary number (and it shouldnot be magic) so its extra lines

mobile/src/libs/components/card/card.tsx
iamAlinaaa: i just wrote iconRight, and @v0ldemar01 suggested me to write Boolean(iconRight) , but it gives an error because we need to check if iconName is not undefined
Снимок экрана 2023-09-28 142701

fshabanov: Did you try `name={iconRight as IconName}`?

dmytro-kucherenko: You can check `iconRight &&` only or in props use assertion like `name={iconRight as iconName}`

iamAlinaaa: I did only iconRight && before , but I was said to change this , so I’ll use this name={iconRight as iconName}

mobile/src/libs/components/button/styles.ts
```diff
@@ -54,6 +57,14 @@ const styles = StyleSheet.create({
   visuallyCenteredButton: {
     paddingLeft: 18,
   },
+  addButtonLabel: {
+    color: AppColor.BLUE_300,
+  },
+  buttonModal: {
+    borderRadius: 8,
+    marginHorizontal: 10,
```
liza-veis: we should avoid margins for common componets
```suggestion
```

mobile/src/screens/chat-list/chat-list.tsx
liza-veis: we should define what is this id for, because it's not obvious. Initial state for items that we get later should be null
```suggestion
  const [chatIdToDelete, setChatIdToDelete] = useState<number | null>(null);
```

mobile/src/screens/chat-list/chat-list.tsx
```diff
@@ -43,6 +47,20 @@ const ChatList: React.FC = () => {
...
   const chatsLength = chats.length;
 
+  const handleShowModal = (id: number): void => {
+    setChatId(id);
+    setIsModalVisible(true);
+  };
+
+  const hanleCloseModal = (): void => {
+    setIsModalVisible(false);
+  };
```
liza-veis: rename this to understand what is this modal for (delete operation)

mobile/src/libs/components/modal/modal.tsx
```diff
@@ -1,28 +1,64 @@
...
+} from '#libs/components/components';
 import { AppColor } from '#libs/enums/enums';
 
 import { styles } from './styles';
 
 type Properties = {
   isVisible: boolean;
   onClose: () => void;
+  onDelete?: () => void;
+  type?: 'Note' | 'Chat';
```
liza-veis: common modal component shouldn't know about this, we can pass buttons here, but we should control this from outside 
```suggestion
```

liza-veis (changes requested):

mobile/src/libs/components/card/card.tsx
liza-veis: FYI you should always use boolean values in conditional rendering (especially in react native and especially with strings) because if we get an empty string for example then this condition will render empty string which is incorrect and will cause an error in react native. Why we should prefer `as iconName` instead of `&& iconName` - because we already checked that icon name exists, so this only ts issue that can be fixed using casting without unnecessary logic

mobile/src/libs/components/modal/modal.tsx
iamAlinaaa: @Liza-Veis our modal component is used for adding meditation entries also, so only when i pass onDelete it is shows functionality for deletion. and we control the deletion from outside for sure, but only pass types to show it in button title

v0ldemar01 (approved):

dmytro-kucherenko (approved):

fshabanov (approved):

mobile/src/libs/components/modal/modal.tsx
liza-veis: we can't have properties like onDelete in modal, because this will mean that we want to delete a modal, also we need to remove type here because modals will have the same look. We should pass common config instead without defining for what functionality it will be used, take a look at how it's passed in react native [alert](https://reactnative.dev/docs/alert#alertbuttonstyle-ios) component

PR #343, opened 2023-09-21
cp-318: Not highlighted icons in the navigation bar

OlesiaTr:
I might have missed in our project, if it is better when the logic is written in one line or divided into separate variables, so let me know if I should change it

https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/104642736/a71818b8-811c-4f6f-8b0a-f723f5e31055


OlenaYatseniuk (dismissed):

frontend/src/libs/components/dropdown-menu/dropdown-menu.tsx
```diff
@@ -35,12 +35,17 @@ const DropdownMenu: React.FC<Properties> = ({ routes }) => {
         )}
       >
         {routes.map((item) => {
+          const isRootPathSelected = pathname === item.path;
+          const isOtherPathSelected =
```
v0ldemar01: Create a common helper for this and use enum as value instead of '/'

v0ldemar01 (changes requested):

frontend/src/libs/components/dropdown-menu/dropdown-menu.tsx
OlesiaTr: Yes, changed it!

frontend/src/libs/helpers/is-route-selected.ts
```diff
@@ -0,0 +1,16 @@
+import { AppRoute } from '#libs/enums/app-route.enum.js';
+
+type Properties = {
+  pathname: string;
+  routePath: string;
+};
+
+const isRouteSelected = ({ pathname, routePath }: Properties): boolean => {
+  const isRootPathSelected = pathname === routePath;
+  const isOtherPathSelected =
```
HIMDanny: maybe let's name it like this?
```suggestion
  const isChildPathSelected =
```

OlesiaTr: Why would it be a child?🤔
As in non root?

HIMDanny: as I understood, root path can be /chats and other path is /chats/2 for example, right?

frontend/src/libs/components/dropdown-menu/dropdown-menu.tsx
```diff
@@ -35,12 +35,17 @@ const DropdownMenu: React.FC<Properties> = ({ routes }) => {
         )}
       >
         {routes.map((item) => {
+          const isSelected = isRouteSelected({
```
v0ldemar01: Prefer quality criteria here

v0ldemar01 (changes requested):

frontend/src/libs/helpers/is-route-selected.ts
HIMDanny: if so, it is not so clear, what is the other path name means. The other path can be /journal for example. So that's why I suggested to name it like isRootChildPathSelected or similar

frontend/src/libs/components/dropdown-menu/dropdown-menu.tsx
OlesiaTr: @v0ldemar01  do you mean better to name it getSelectedRoute?

frontend/src/libs/helpers/is-route-selected.ts
OlesiaTr: Yes, that makes sense!

frontend/src/libs/components/dropdown-menu/dropdown-menu.tsx
dmytro-kucherenko: I think naming is good at now, maybe `getIsSelectedRoute` will be better option

dmytro-kucherenko (dismissed):

frontend/src/libs/helpers/get-selected-route.ts
```diff
@@ -0,0 +1,16 @@
+import { AppRoute } from '#libs/enums/app-route.enum.js';
+
+type Properties = {
+  pathname: string;
+  routePath: string;
+};
+
+const getSelectedRoute = ({ pathname, routePath }: Properties): boolean => {
```
liza-veis: I think this is not a good approach, we should pass needed routes explicitly in routes config

liza-veis (changes requested):

OlesiaTr:
@Liza-Veis sorry, I missed the comment. Could you please confirm if this is the approach you had in mind, or if there are any further suggestions or adjustments you'd like me to make?

frontend/src/libs/components/dropdown-menu/dropdown-menu.tsx
v0ldemar01: getIsSelectedRoute => checkIsSelectedRoute

OlesiaTr: ohhh ok, will change it

frontend/src/libs/helpers/check-is-selected-route.ts
```diff
@@ -0,0 +1,20 @@
+import { AppRoute } from '#libs/enums/app-route.enum.js';
```
liza-veis: we should fully remove this helper and check it another way. I meant that the idea of checking by `startsWith` isn't good, we should strictly compare routes, that's why we need to pass it to config somehow

liza-veis (changes requested):

frontend/src/libs/helpers/check-is-selected-route.ts
OlesiaTr: @Liza-Veis , I'm sorry, I forgot to check for a response again. I couldn't find a route config? The only thing I found is SIDEBAR_ROUTES, but that's what we are passing as routes to the Navigation menu. Can you give me a bit details, so I could close it today already?

v0ldemar01 (approved):

PR #348, opened 2023-09-21
cp-315: Modify response body for journal requests

dmytro-kucherenko:
Done: 
- Add to swagger documentation response for the case when the length of the title is over the limit (limit is 400 chars) (the only case was journal creating)

All other mentioned problems in the issue was resolved in pr #293 

image

### DIFF
--- a/backend/src/packages/journal-entries/journal-entry.controller.ts
+++ b/backend/src/packages/journal-entries/journal-entry.controller.ts
@@ -77,6 +77,9 @@ class JournalEntryController extends BaseController {
     this.addRoute({
       path: JournalApiPath.ROOT,
       method: 'POST',
+      validation: {
+        body: createJournalEntryValidationSchema,
+      },
       handler: (options) => {
         return this.create(
           options as APIHandlerOptions<{
@@ -173,6 +176,12 @@ class JournalEntryController extends BaseController {
    *              example:
    *                message: "User with these credentials was not found."
    *                errorType: "USERS"
+   *        422:
+   *          description: Validation failed
+   *          content:
+   *            application/json:
+   *              schema:
+   *                $ref: '#/components/schemas/Unprocessable Journal Entity'
    */
 
   private async create(


backend/src/packages/journal-entries/journal-entry.controller.ts
```diff
@@ -77,6 +77,9 @@ class JournalEntryController extends BaseController {
     this.addRoute({
       path: JournalApiPath.ROOT,
       method: 'POST',
+      validation: {
+        body: createJournalEntryValidationSchema,
+      },
```
OlenaYatseniuk, line 82: It is unnecessary here, as in create I just send the default payload, so user can't send anything
```suggestion
      
```

backend/src/packages/journal-entries/journal-entry.controller.ts
```diff
@@ -173,6 +176,12 @@ class JournalEntryController extends BaseController {
    *              example:
    *                message: "User with these credentials was not found."
    *                errorType: "USERS"
+   *        422:
+   *          description: Validation failed
+   *          content:
+   *            application/json:
+   *              schema:
+   *                $ref: '#/components/schemas/Unprocessable Journal Entity'
```
OlenaYatseniuk, line 184:
```suggestion
   *        422:
   *          description: Validation failed
   *          content:
   *            application/json:
   *              schema:
   *                $ref: '#/components/schemas/Unprocessable Journal Entity'
```
So here it is impossible to have this error too

backend/src/packages/journal-entries/journal-entry.controller.ts
fshabanov: @OlenaYatseniuk it's always good to add a validation in backend, because the user might use Swagger or Postman to make requests

fshabanov (approved):

v0ldemar01 (approved):

backend/src/packages/journal-entries/journal-entry.controller.ts
OlenaYatseniuk: @fshabanov ok, thanks for the clarification

OlenaYatseniuk (approved):

BrianSammit (approved):

danilittaa (approved):

backend/src/packages/journal-entries/journal-entry.controller.ts
dmytro-kucherenko: Yeah, I was also taking into consideration a Postman when was adding it

iamAlinaaa (approved):

liza-veis (approved):

PR #360, opened 2023-09-22
cp-282: Load Mobile Meditation Entries

iamAlinaaa:
when user clicks on meditation item on bottom navigation bar, all meditation entries are loaded

Снимок экрана 2023-09-22 124057

### DIFF
--- a/mobile/src/packages/meditation/libs/types/types.ts
+++ b/mobile/src/packages/meditation/libs/types/types.ts
@@ -1,0 +1,5 @@
+export {
+  type MeditationEntryCreateForm,
+  type MeditationEntryCreateRequestDto,
+  type MeditationEntryCreateResponseDto,
+} from 'shared/build/index.js';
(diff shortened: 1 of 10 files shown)


mobile/src/packages/meditation/meditation-api.ts
```diff
@@ -0,0 +1,57 @@
...
+      {
+        method: 'POST',
+        payload: formData,
+        hasAuth: true,
+      },
+    );
+
+    return await response.json<MeditationEntryCreateResponseDto>();
+  }
+
```
danilittaa: Why do we need this function in this pr?

mobile/src/packages/meditation/libs/types/types.ts
```diff
@@ -0,0 +1,7 @@
+export {
+  type MeditationEntryCreateForm,
+  type MeditationEntryCreateRequestDto,
+  type MeditationEntryCreateResponseDto,
```
dmytro-kucherenko, line 4: The same, we don't need create dtos here for now

iamAlinaaa: @Dmytro-Kucherenko they are needed because this type is used in my task component
Снимок экрана 2023-09-22 155629

liza-veis (approved):

v0ldemar01 (approved):

PR #362, opened 2023-09-22
cp-342: Modify response for meditation post

OlesiaTr:
- [x]  Added 'id' to the response for 201 status code
- [x]  Added to swagger documentation sample responses for incorrect format and for file size over 10Mb.

adding_id

### DIFF
--- a/backend/src/packages/meditation/meditation.service.ts
+++ b/backend/src/packages/meditation/meditation.service.ts
@@ -43,10 +43,9 @@ class MeditationService implements Service {
     name: string;
     file: FileUploadRequestDto;
   }): Promise<MeditationEntryCreateResponseDto> {
-    const { id, url, contentType } = await this.fileService.create(file);
+    const { url, contentType } = await this.fileService.create(file);
     const item = await this.meditationRepository.create(
       MeditationEntity.initializeNew({
-        id,
         name,
         mediaUrl: url,
         contentType,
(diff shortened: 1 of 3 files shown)


liza-veis (dismissed):

backend/src/packages/meditation/meditation.service.ts
```diff
@@ -43,10 +43,11 @@ class MeditationService implements Service {
     name: string;
     file: FileUploadRequestDto;
   }): Promise<MeditationEntryCreateResponseDto> {
-    const { url, contentType } = await this.fileService.create(file);
+    const { id, url, contentType } = await this.fileService.create(file);
     const item = await this.meditationRepository.create(
       MeditationEntity.initializeNew({
-        name: name,
+        id,
```
fshabanov: Why are you using `id` of the file to create a meditation?

fshabanov (changes requested):

OlenaYatseniuk (commented):
Can you also add, please, security to post route as it is in get route?

backend/src/packages/meditation/meditation.service.ts
OlesiaTr: Because I thought they should be added to the meditation entries

liza-veis: Farid is right, id's are created automatically, you task is to return it in response not to set it

liza-veis (changes requested):

backend/src/packages/meditation/meditation.service.ts
OlesiaTr: Yes, sorry, you are correct, I got it misunderstood

liza-veis (approved):

fshabanov (approved):

HIMDanny (approved):

PR #365, opened 2023-09-22
cp-346: Sort chats by newest messages

OlenaYatseniuk:
https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/102380564/3306a7aa-d595-48ea-843a-a0c972400a30


backend/src/packages/chats/chat.repository.ts
```diff
@@ -70,8 +74,8 @@ class ChatRepository implements Repository {
     });
   }
 
-  public update(): Promise<unknown> {
-    return Promise.resolve(null);
+  public async update({ id, updatedAt }: UpdateChatPayload): Promise<void> {
+    await this.chatModel.query().patchAndFetchById(id, { updatedAt });
   }
```
v0ldemar01:
```suggestion
  publiс update({ id, updatedAt }: UpdateChatPayload): Promise<ChatEntity> {
    return this.chatModel.query().patchAndFetchById(id, { updatedAt }).execute();
  }
```

backend/src/packages/chats/chat.service.ts
```diff
@@ -73,10 +73,19 @@ class ChatService implements Service {
...
   ): Promise<ChatMessageGetAllItemResponseDto> {
-    return this.chatMessageService.create(payload);
+    const createdMessage = await this.chatMessageService.create(payload);
+
+    const { chatId, updatedAt } = createdMessage;
+
+    await this.chatRepository.update({
+      id: chatId,
+      updatedAt: updatedAt.toDateString(),
+    });
```
v0ldemar01: Not sure whether it is needed

v0ldemar01 (changes requested):

backend/src/packages/chats/chat.service.ts
OlenaYatseniuk: @v0ldemar01 Because for now, chat doesn't update if there is a new message in it, and I am sure that it is incorrect. If I did this, I update its updatetAt, so I can send all chats in the DESC way and this is a requirement for this task - "Chats with the newest messages should be shown on top of the chat list"

backend/src/packages/chats/chat.repository.ts
OlenaYatseniuk: @v0ldemar01 But this method doesn't return ChatEntity, it returns ChatModel. Actually, it isn't needed, I guess, as I want only to update updatedAt column in chat, I don't need any return value to use then. If I  am wrong, please, correct me.

v0ldemar01: Yep, ChatModel. But if you are using pathAndFetch then the function will return result

backend/src/packages/chats/chat.service.ts
v0ldemar01: You can just sort using db query without setting new value to updatedAt

OlenaYatseniuk: @v0ldemar01 But how can I sort it, if updatetAt never updates?
Sort without it just send chats in order when they were created, not updated by a new message.

iamAlinaaa (dismissed):

backend/src/packages/chats/chat.service.ts
v0ldemar01: Just make a join with messages and sort by newest message

OlenaYatseniuk: Will try to reach it

v0ldemar01 (approved):

liza-veis (approved):

PR #345, opened 2023-09-21
cp-276: Group chat messages by date

dmytro-kucherenko:
Done:
- chat date divider should be relative

image

PR #377, opened 2023-09-23
cp-369: Display Time Near Chat Messages

danilittaa:
image


mobile/src/libs/helpers/message-item-properties.ts
```diff
@@ -0,0 +1,58 @@
...
+};
+
+type MessageItemProperties = {
+  isTimeVisible: boolean;
+  isDifferentMessageOwner?: boolean;
+  currentTime?: string;
+};
+
+const PREVIOUS_USER = 1;
+const NEXT_USER = 1;
```
iamAlinaaa: Maybe its better to move constants to separate file ?

OlesiaTr (dismissed):

mobile/src/screens/chat/libs/helpers/message-item-properties.helper.ts
```diff
@@ -0,0 +1,58 @@
...
+  currentIndex: number;
+};
+
+type MessageItemProperties = {
+  isTimeVisible: boolean;
+  isDifferentMessageOwner?: boolean;
+  currentTime?: string;
+};
+
+const messageItemProperties = ({
```
v0ldemar01: I suppose we can group messages on backend side https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/pull/345

v0ldemar01 (changes requested):

mobile/src/screens/chat/libs/helpers/message-item-properties.helper.ts
danilittaa: Do you mean asking to add time grouping as well? In general, it might not make much sense because as soon as I send a message, I immediately receive a response from the chatbot. 

So, we can even just display the time next to the message. I just decided to add this functionality just in case because I remember the product owner said to split the bot's messages into several small ones rather than sending one big message, but we won't have time to implement it anymore. So, I think it would be better to either completely remove this functionality or leave it on the frontend only without interacting with the backend, as the logic here is not too complex. What do you think?

mobile/src/screens/chat/chat.tsx
```diff
@@ -101,16 +102,19 @@ const Chat: React.FC = () => {
...
+            });
 
           return (
             <MessageItem
               text={item.message}
+              time={currentTime ?? '00:00'}
+              isTimeVisible={isTimeVisible}
               isUser={item.senderId === authenticatedUser.id}
-              isAvatarVisible={isDifferentMessageOwner}
+              isAvatarVisible={isDifferentMessageOwner ?? true}
```
dmytro-kucherenko: Is `isDifferentMessageOwner` a `boolean | undefined`? In any case `?? true` is quite hard to understand from the first view

mobile/src/screens/chat/libs/helpers/message-item-properties.helper.ts
dmytro-kucherenko: In my opinion we are able not to implement it for now because of bot can't send more than one message

liza-veis (dismissed):

liza-veis (approved):

PR #384, opened 2023-09-24
cp-170: Add Meditation Player Layout

danilittaa:
- fixed styles
- player stops when the user navigates to another page, turns off the screen, or exits the application


mobile/src/libs/helpers/get-formatted-time/get-formatted-time.helper.ts
v0ldemar01, line 1: We need to figure out how to accommodate that helper

v0ldemar01 (changes requested):

mobile/src/libs/helpers/get-formatted-time/get-formatted-time.helper.ts
danilittaa: Could you please explain what you mean?

v0ldemar01: We should use this one for frontend and mobile indeed. So Looks like the issue is related to timezone of prod mode.

danilittaa: But I really don't need to take into account the time zone, as it simply displays the track's time. If I don't understand something, please correct me

v0ldemar01: The app can be deployed into any timezone, so we need to in touch with it

OlesiaTr (dismissed):

iamAlinaaa (dismissed):

mobile/src/libs/components/player/components/controls/controls.tsx
```diff
@@ -25,6 +30,12 @@ const Controls: React.FC<Properties> = ({ isPlaying: initialIsPlaying }) => {
     handlePlayPause();
   };
 
+  useFocusEffect(
+    useCallback(() => {
+      setIsPlaying(false);
+    }, []),
```
dmytro-kucherenko: Shouldn't `setIsPlaying` be in dependencies? The same with dependencies of `useCallback`'s in other places

mobile/src/libs/components/player/player.tsx
```diff
@@ -61,6 +67,33 @@ const Player: React.FC<Properties> = ({ setCurrentTrack }) => {
     setCurrentTrack(MOCKED_PLAYLIST[TRACK_START_INDEX] as Track);
   }, [setCurrentTrack]);
 
+  useEffect(() => {
+    const handleAppStateChange = (nextAppState: string): void => {
+      void player.startPlayling();
+
+      if (nextAppState === 'background') {
```
dmytro-kucherenko: May we create `AppState` enum for that reason?

mobile/src/libs/helpers/get-formatted-time/get-formatted-time.helper.ts
dmytro-kucherenko: We have helper with the same name in shared folder, is they're the different things?

v0ldemar01 (approved):

mobile/src/libs/packages/player/player.package.ts
```diff
@@ -39,6 +39,10 @@ class Player {
     await TrackPlayer.pause();
   };
 
+  public startPlayling = async (): Promise<void> => {
```
v0ldemar01:
```suggestion
  public startPlaying = async (): Promise<void> => {
```

v0ldemar01 (changes requested):
Small typo issue

v0ldemar01 (approved):

fshabanov (approved):

PR #393, opened 2023-09-25
cp-337: Make plus buttons identical

Xxandder:
- Made plus buttons identical on chat and meditation list sidebard

image
image


frontend/src/assets/css/scaffolding.scss
```diff
@@ -14,6 +14,10 @@ body {
   line-height: 1.5;
 }
 
+button {
+  padding: 0;
+}
+
```
liza-veis: remove this, reset padding in link component
```suggestion
```

frontend/src/pages/chat/components/chat-sidebar/styles.module.scss
```diff
@@ -12,6 +13,12 @@
   border-radius: 24px;
 }
 
+.plus {
+  display: flex;
+  height: 33px;
+  margin-left: auto;
+}
```
liza-veis:
```suggestion
```

liza-veis (changes requested):

frontend/src/pages/chat/components/chat-sidebar/styles.module.scss
Xxandder: When I remove height or max-height here I give the next result and there is no paddings:
image
image

Xxandder: @Liza-Veis May I leave height or max-height property?

liza-veis: no, because this mean that layout is incorrect, button itself should take this space and we should have the same styles for button in both places

liza-veis (approved):

PR #394, opened 2023-09-25
cp-356: make height of sidebars headers identical

Xxandder:
- Made sidebars height identical. 
In chat and meditation pages we have plus icons with 33px height. So that is why I set max-height property to make sidebars headers look identical.
image
image
image


Xxandder:
Icons sizes are different in this pr, but I have also pushed a pr with setting plus icons sizes identical

frontend/src/pages/chat/components/chat-sidebar/styles.module.scss
```diff
@@ -2,6 +2,7 @@
   display: flex;
   justify-content: center;
   align-items: center;
+  max-height: 30px;
```
liza-veis: remove this everywhere, this should be done without it
```suggestion
```

liza-veis (changes requested):

frontend/src/pages/chat/components/chat-sidebar/styles.module.scss
liza-veis: we should consider case where text will take two lines, and header height will be different in this case

Xxandder: Ok, understood. I just thought it's ok to have less paddings in case when header has two rows

v0ldemar01 (approved):

vladyslavdobychin (approved):

liza-veis (approved):

Issue #370, opened 2023-09-23
Change color for the fonts throughout the application


oksana-mlynska:
Change color for the fonts throughout the application from #000000 to  #302E36

![Image](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/140756353/26da7627-d7ef-4b2a-9768-61463c6f147b)

I see that we have it set like grey 400

![Image](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/140756353/0c738e34-31f2-4d94-9b1d-b60cd1f96e0f)

PR #399, opened 2023-09-25
cp-370: Change font color

vladyslavdobychin:
As was mentioned in the [issue](https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/issues/370) changed all instances of color #000000 to #302E36 (--gray-400)


Ekit18 (dismissed):

OlenaYatseniuk (dismissed):

fshabanov:
@vladyslavdobychin have you tried to update the default text color instead of setting the new color everywhere?

vladyslavdobychin:
> @vladyslavdobychin have you tried to update the default text color instead of setting the new color everywhere?

I've been thinking about that, but I saw multiple occasion of text being colored differently, like text of a button or text inside of the add file modal. I was just trying to avoid making changes somewhere I wasn't supposed to.

vladyslavdobychin:
@fshabanov and how would you advise to do this?

liza-veis:
We should set it in global styles, but also for such components as input, button, textarea we should set it one more time

liza-veis (approved):

v0ldemar01 (approved):

PR #401, opened 2023-09-25
cp-395: Remove Focus On Input When Keyboard Is Closed

danilittaa:
The problem was that when the user pressed the back button (thus closing the keyboard), the focus did not shift away from the input field. As a result, when the user pressed the 'send message' button, the keyboard would activate again


fshabanov (approved):

ossavemu (approved):
Well approach!

dmytro-kucherenko (approved):

iamAlinaaa (approved):

PR #406, opened 2023-09-26
cp-257: Fixing responsive height in the side menu

juanc-escobar:
Fixed media queries to be responsive to height.

height responsive


juanc-escobar:
@Liza-Veis 

ci-pr Expected — Waiting for status to be reported. 

this message won't update and does not show Details.

PR #417, opened 2023-09-26
cp-277: Group Chat Messages By Date

danilittaa:
- messages are visually separated based on their relative date

image


iamAlinaaa (commented):
I think there should be some more space / padding between grey line of date and message

shared/src/libs/helpers/date/get-relative-date/get-relative-date.helper.ts
```diff
@@ -0,0 +1,21 @@
+import { formatRelative } from 'date-fns';
+import enGB from 'date-fns/locale/en-GB/index.js';
+
+import { type ValueOf } from '#libs/types/types.js';
+
+import { FormatRelativeLocale } from '../libs/enums/enums.js';
+
+const getRelativeDate = (date: Date): string => {
```
v0ldemar01: Please check related PR for a web part

v0ldemar01 (changes requested):

mobile/src/screens/chat/chat.tsx
```diff
@@ -49,7 +49,8 @@ const Chat: React.FC = () => {
   );
 
   const scrollViewReference = useRef<ScrollView | null>(null);
-  const messagesLength = currentChatMessages.length;
+  const messagesLength = Object.keys(currentChatMessages).length;
```
dmytro-kucherenko:
```suggestion
  const messagesLength = Object.keys(currentChatMessages).flat().length;
```

I also changed that in my pr, cause before it was only the count of days when messages were sent

shared/src/libs/helpers/date/libs/enums/format-relative-locale.enum.ts
```diff
@@ -0,0 +1,13 @@
+import { SINGLE_QUOTE_UNICODE } from '../constants/constants.js';
```
dmytro-kucherenko: The approach in helper was changed, so you can delete this enum at all

shared/src/libs/helpers/date/libs/constants/constants.ts
```diff
@@ -1,4 +1,6 @@
 const MILLISECONDS_IN_SECOND = 1000;
 const SECONDS_IN_MINUTE = 60;
 
-export { MILLISECONDS_IN_SECOND, SECONDS_IN_MINUTE };
+const SINGLE_QUOTE_UNICODE = '\u0027';
```
dmytro-kucherenko: The same with that constant

mobile/src/screens/chat/chat.tsx
danilittaa: Thank you! I wasnt attentive

liza-veis (approved):

v0ldemar01 (approved):

PR #424, opened 2023-09-26
cp-421: Improved Mobile Meditation Player

iamAlinaaa:
1. Mobile Meditation player progress bar improvement
2. Checked the color of all requested items, it is GRAY_400: "#ACB4D0"
3. Reduced font size of timestamp

Снимок экрана 2023-09-27 101650


fshabanov (dismissed):

danilittaa (commented):
Why did you only complete the first item instead of all three, as described in the ticket?

iamAlinaaa:
> Why did you only complete the first item instead of all three, as described in the ticket?

the colors were the same everywhere and timestamp seemed quite small even without reducing, but i reduced its size and added color as required in ticket now

danilittaa (approved):
It looks good now)

v0ldemar01 (approved):

dmytro-kucherenko (approved):

liza-veis (approved):

PR #447, opened 2023-09-27
cp-78: Fix layout bugs in meditation player

OlenaYatseniuk:
- fix all layout bugs in the meditation player
- wrap card in meditation sidebar by Link
- add a back button to the meditation player
- reduce progress bar and controls button width

https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/102380564/71f8d1f0-7769-4410-b475-3d6e27bd1ee7


frontend/src/libs/components/backward-button/backward-button.tsx
```diff
@@ -1,15 +1,23 @@
 import { IconColor } from '#libs/enums/enums.js';
+import { getValidClassNames } from '#libs/helpers/helpers.js';
 
 import { Icon } from '../components.js';
 import styles from './styles.module.scss';
 
 type Properties = {
   onGoBack: () => void;
+  isVisible?: boolean;
```
v0ldemar01: it is useless

v0ldemar01 (changes requested):

frontend/src/libs/components/backward-button/backward-button.tsx
OlenaYatseniuk: Well, I rewrote the styles of the back button, and as we need this arrow on the big screen only for the meditation player, so I need this

v0ldemar01: But we need to handle the visibility outside the button

OlenaYatseniuk: Well, I need to change the left position for the button if it is a desktop, and if I return back the style of the back button where it has display - none in desktop resolution, how could I handle it in meditation player component if I want to have it visible?

v0ldemar01: Just provide visibility condition to wrapper around button, follow single responsibility

frontend/src/libs/components/backward-button/back-button-wrapper.tsx
```diff
@@ -0,0 +1,21 @@
...
+  onGoBack: () => void;
+  isVisible?: boolean;
+};
+
+const BackButtonWrapper: React.FC<BackButtonWrapperProperties> = ({
+  isVisible = false,
+  onGoBack,
+}) => {
+  return (
+    <div className={getValidClassNames(!isVisible && 'visually-hidden')}>
```
v0ldemar01: Cannot we add 'visually-hidden' for <div> wrapper over BackButtonWrapper

frontend/src/pages/meditation/components/meditation-player/meditation-player.tsx
```diff
@@ -22,11 +26,14 @@ const MeditationPlayer: React.FC = () => {
...
   const { id: meditationEntryId } = useParams<{ id: string }>();
   const [searchParameters] = useSearchParams();
+  const { setIsSidebarShow } = useSidebarState();
+  const navigate = useNavigate();
   const timerDuration = Number(
     searchParameters.get(AppQueryStringKey.TIMER_DURATION),
   );
 
   const [trackIndex, setTrackIndex] = useState(TRACK_FIRST_INDEX);
+  const isMobileDimension = useMediaQuery(MOBILE_DIMENSION);
```
v0ldemar01, line 36: Cannot we handle it with css media queries?

v0ldemar01 (changes requested):

frontend/src/pages/meditation/components/meditation-player/meditation-player.tsx
OlenaYatseniuk: I can put something like that: `    const handleBackButtonPress = useCallback(() => {
    navigate(AppRoute.MEDITATION);
    if (window.innerWidth < 768) {
      setIsSidebarShow(false);
    }
  }, [navigate, setIsSidebarShow])`
  But for me, existing implementation is better.

OlenaYatseniuk: I can't do this only by scss, as I need to know whether it needs to add "?sidebarMode=show" to URL, when I click on this button. Because in case it is a list in desktop dimension I don't need this button, if it is mobile I need it, if I don't add "?sidebarMode=show" in mobile I can't see this list component.
So, yesterday I tried to deal with it with scss, but it doesn't work as expected.

frontend/src/libs/components/backward-button/back-button-wrapper.tsx
OlenaYatseniuk: No, we can't. As here we have div (container) and has its own functionality. Because in meditation list we don't need to have this back button in desktop, but the container is needed. So I can't move this logic to it. That's why I created this BackButtonWrapper. 
image

OlenaYatseniuk: I can delete this component and just do the same with the usual div there, but as I need this in several components I decided to make it as a component to reduce the repetition.

v0ldemar01: I mean to wrap BackButtonWrapper with separate <div></div>

OlenaYatseniuk: I don't understand why we need one extra div here, as I have it in this BackButtonWrapper , which I created exactly for this purpose. Could you explain this, please?

v0ldemar01: Sorry, I missed that, let's proceed with your solution

v0ldemar01 (dismissed):

frontend/src/libs/components/components.ts
```diff
@@ -1,6 +1,7 @@
 export { App } from './app/app.js';
 export { AudioPlayer } from './audio-player/audio-player.js';
-export { BackButton } from './backward-button/backward-button.js';
+export { BackButtonWrapper } from './backward-button/back-button-wrapper.js';
+export { BackButton } from './backward-button/components/backward-button.js';
```
fshabanov: Name of the file and component should be the same
Don't forget about folder name

frontend/src/libs/hooks/hooks.ts
```diff
@@ -2,6 +2,7 @@ export { useAppDispatch } from './use-app-dispatch/use-app-dispatch.hook.js';
 export { useAppForm } from './use-app-form/use-app-form.hook.js';
 export { useAppSelector } from './use-app-selector/use-app-selector.hook.js';
 export { useHandleClickOutside } from './use-handle-click-outside/use-handle-click-outside.hook.js';
+export { useMediaQuery } from './use-media-query/use-media-query.js';
```
fshabanov:
```suggestion
export { useMediaQuery } from './use-media-query/use-media-query.hook.js';
```

frontend/src/libs/hooks/use-media-query/use-media-query.ts
```diff
@@ -0,0 +1,27 @@
+import { useCallback, useEffect, useState } from '../hooks.js';
+
+const useMediaQuery = (query: string): boolean => {
+  const [matches, setMatches] = useState(false);
```
fshabanov: Rule B19 about boolean naming

fshabanov (changes requested):

frontend/src/libs/components/components.ts
OlenaYatseniuk: Thanks for your attention!

frontend/src/libs/components/back-button/back-button-wrapper.tsx
```diff
@@ -0,0 +1,21 @@
...
+};
+
+const BackButtonWrapper: React.FC<BackButtonWrapperProperties> = ({
+  isVisible = false,
+  onGoBack,
+}) => {
+  return (
+    <div className={getValidClassNames(!isVisible && 'visually-hidden')}>
+      <BackButton onGoBack={onGoBack} />
+    </div>
```
dmytro-kucherenko, line 17: Maybe it's better to implement this functionality directly in `BackButton` component and set default to be visually displayed. Seems like redundant component for now

frontend/src/libs/constants/index.constant.ts
```diff
@@ -1,4 +1,6 @@
 const FIRST_INDEX = 0;
 const LAST_INDEX = -1;
 
-export { FIRST_INDEX, LAST_INDEX };
+const MOBILE_DIMENSION = '(max-width: 767px)';
```
dmytro-kucherenko: It's the `index.constant.ts` file only, so move it to new one

frontend/src/libs/hooks/use-media-query/use-media-query.hook.ts
```diff
@@ -0,0 +1,27 @@
+import { useCallback, useEffect, useState } from '../hooks.js';
+
+const useMediaQuery = (query: string): boolean => {
+  const [isMatched, setIsMatched] = useState(false);
+
+  const handleMediaQueryChange = useCallback(
+    (event: MediaQueryListEvent): void => {
+      setIsMatched(event.matches);
+    },
+    [],
```
dmytro-kucherenko, line 10: Is it acceptable to have no dependencies there even without `setIsMatched`?

frontend/src/libs/hooks/use-media-query/use-media-query.hook.ts
```diff
@@ -0,0 +1,27 @@
...
+  );
+
+  useEffect(() => {
+    const mediaQueryList = window.matchMedia(query);
+
+    mediaQueryList.addEventListener('change', handleMediaQueryChange);
+
+    setIsMatched(mediaQueryList.matches);
```
dmytro-kucherenko:
```suggestion
    const mediaQueryList = window.matchMedia(query);
    setIsMatched(mediaQueryList.matches);

    mediaQueryList.addEventListener('change', handleMediaQueryChange);
```

If it won't break the logic

frontend/src/pages/meditation/components/meditation-player/meditation-player.tsx
```diff
@@ -1,12 +1,16 @@
 import deafultMeditationImage from '#assets/img/meditation-image-placeholder.jpg';
-import { AudioPlayer } from '#libs/components/components.js';
-import { AppQueryStringKey } from '#libs/enums/enums.js';
+import { AudioPlayer, BackButtonWrapper } from '#libs/components/components.js';
+import { MOBILE_DIMENSION } from '#libs/constants/index.constant.js';
```
dmytro-kucherenko:
```suggestion
import { MOBILE_DIMENSION } from '#libs/constants/constants.js';
```

frontend/src/pages/user-profile/user-profile.tsx
```diff
@@ -1,13 +1,19 @@
-import { BackButton } from '#libs/components/components.js';
+import { BackButtonWrapper } from '#libs/components/components.js';
+import { MOBILE_DIMENSION } from '#libs/constants/index.constant.js';
```
dmytro-kucherenko: The same

frontend/src/libs/components/back-button/back-button-wrapper.tsx
OlenaYatseniuk: You can see comment from @v0ldemar01 that requested to do this

frontend/src/libs/hooks/use-media-query/use-media-query.hook.ts
HIMDanny: I guess React knows that `setIsMatched` would never be recreated and doesn't ask us to add this function in the dependency array

frontend/src/libs/components/back-button/back-button-wrapper.tsx
```diff
@@ -0,0 +1,21 @@
+import { getValidClassNames } from '#libs/helpers/helpers.js';
+
+import { BackButton } from './back-button.js';
+
+type BackButtonWrapperProperties = {
```
fshabanov:
```suggestion
type Properties = {
```

fshabanov (changes requested):

fshabanov (approved):

liza-veis (approved):

PR #448, opened 2023-09-27
cp-423: improvements in dropdown  burger menu

Xxandder:
- Fixed the wrong view of home item when it was active
- Increased the gap between icons and text

image


frontend/src/libs/components/dropdown-menu/styles.module.scss
```diff
@@ -74,11 +74,6 @@
     brightness(93%) contrast(86%);
 }
 
-.selected {
-  background-color: var(--gray-100);
-  cursor: pointer;
-}
-
```
dmytro-kucherenko: Should the selected option highlighting be removed totally? I think user have to see that and the problem was with different `pathname` due to the url params

v0ldemar01: Good point

Xxandder: Yep, thanks for the noticing. I will do it in this way

fshabanov (dismissed):

frontend/src/libs/components/dropdown-menu/dropdown-menu.tsx
```diff
@@ -11,15 +12,28 @@ type Properties = {
...
 
+  const checkIsCurrentRoot = useCallback(
+    (currentPath: string, itemPath: string): boolean => {
+      const currentPathWithoutSlash = currentPath.slice(SECOND_INDEX);
+      const itemPathWithoutSlash = itemPath.slice(SECOND_INDEX);
+
+      return (
+        currentPath === itemPath ||
+        (currentPathWithoutSlash !== '' &&
+          itemPathWithoutSlash.startsWith(currentPathWithoutSlash))
```
v0ldemar01: We already have helper for it for sidebar

v0ldemar01 (changes requested):

OlenaYatseniuk (approved):

liza-veis (approved):

PR #454, opened 2023-09-27
cp-431: Fix search request

Ekit18:
https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/92929504/205712f3-7fd0-455e-8ac8-0364d7d08fb4


frontend/src/libs/enums/app-route.enum.ts
```diff
@@ -9,6 +9,8 @@ const AppRoute = {
   CHATS_$ID: '/chats/:id',
   JOURNAL: '/journal',
   JOURNAL_$ID: '/journal/:id',
+  JOURNAL_$ID_QUERY: '/journal/:id?query=:query',
+  CHATS_$ID_QUERY: '/chats/:id?query=:query',
```
v0ldemar01: It's not a good idea

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
```diff
@@ -68,7 +68,8 @@ const ChatLayout: React.FC<Properties> = ({ filter }) => {
     if (createMessageDataStatus === DataStatus.FULFILLED) {
       void dispatch(chatActions.getAllChats(filter));
     }
-  }, [createMessageDataStatus, dispatch, filter]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
v0ldemar01: Not completely sure that is needed

v0ldemar01 (changes requested):

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
Ekit18: @v0ldemar01 i deleted filter from dependencies to avoid 2 request to backend, because filter in dependencies is useless in this case. but without commenting eslint rule i can`t do that

frontend/src/libs/enums/app-route.enum.ts
Ekit18: @v0ldemar01 without that we have this bug:

https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/92929504/6e45a440-42de-417c-b5bb-f8650746987d

v0ldemar01: Got it, but it's not a good practise for us

v0ldemar01: Try to figure out the another one

v0ldemar01 (changes requested):

frontend/src/libs/enums/app-route.enum.ts
Ekit18: @v0ldemar01 but we use link component to open chat/journal where i can pass only variables from app-route.enum.ts. so i should change it or may be leave?

Ekit18: @v0ldemar01 i find only two variant: leave it and add to router-provider/libs/maps/maps.tsx Partial type, or add to link component additional type. I can`t find another variant without modifying url of the link in sidebar

frontend/src/pages/journal/journal.tsx
```diff
@@ -37,23 +33,14 @@ const Journal: React.FC = () => {
...
-  }, [handleGetSelectedNote, id]);
-
   return (
     <>
       <JournalSidebar
         isSidebarShown={isSidebarShown}
         onSetIsSidebarShow={setIsSidebarShow}
         filter={filter}
         onSetFilter={setFilter}
+        id={id}
```
liza-veis: don't pass id, this should be passed in functions how it was before in this component
```suggestion
```

frontend/src/libs/components/link/link.tsx
```diff
@@ -6,7 +6,7 @@ import { type ValueOf } from '#libs/types/types.js';
 import styles from './styles.module.scss';
 
 type Properties = {
-  to: ValueOf<typeof AppRoute>;
+  to: ValueOf<typeof AppRoute> | `${ValueOf<typeof AppRoute>}?query=${string}`;
```
liza-veis:
```suggestion
  to: ValueOf<typeof AppRoute>;
```

frontend/src/pages/chat/components/chat-sidebar/chat-sidebar.tsx
```diff
@@ -104,7 +104,7 @@ const ChatSidebar: React.FC<Properties> = ({
               ) as ValueOf<typeof AppRoute>;
 
               return (
-                <Link key={chat.id} to={chatLink}>
+                <Link key={chat.id} to={`${chatLink}?query=${filter}`}>
```
liza-veis: use helper getUrlWithQueryString as I remember

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
liza-veis: don't we need to getAllMessages on filter change? or how does this work

liza-veis (changes requested):

frontend/src/libs/components/link/link.tsx
Ekit18: @Liza-Veis without it i can`t pass url with query params

Ekit18:
@Liza-Veis in this case when we create message for chat-bot instead of adding message, which comes from backend to store, we rewrite our store, invoking method getAllChats(filter) to get new message on frontend when createMessageState becomes fullfiled. But it's placed in useEffect, and i invoke same method in chat sidebar when i change filter because  createMessageState is fullfiled. Also i pass to this method filter, and if i don`t remove filter from useEffect dependecies in chat layout it will invoke twice when user write smth in search field, because both getAllChats(filter) methods are placed in useEffect and have in dependecies filter

frontend/src/libs/components/link/link.tsx
Ekit18: done

frontend/src/pages/journal/journal.tsx
Ekit18: done

frontend/src/pages/chat/components/chat-sidebar/chat-sidebar.tsx
Ekit18: done

liza-veis:
can't we call getAllChats in one place in this case? or fix the logic for the second one

v0ldemar01 (approved):

PR #461, opened 2023-09-27
cp-446: Add loader to chatbot messages

HIMDanny:
- Added a loader when chatbot response is loading
 - User is not allowed to send something in the chat during this time
 - Added a notification why user can't send a message

### Response is loading
image

### Notification
image


frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
```diff
@@ -86,9 +92,17 @@ const ChatLayout: React.FC<Properties> = ({ filter }) => {
               />
             );
           })}
+        {isChatbotReplyLoading && (
+          <div>
```
liza-veis: do we need this div around loader?

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
```diff
@@ -86,9 +92,17 @@ const ChatLayout: React.FC<Properties> = ({ filter }) => {
...
+            <span className={styles['loader']} />
```
liza-veis: why can't we use our loader component?

liza-veis (changes requested):

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
HIMDanny: we can't, because it has a position absolute so it would overlap the whole page. To use our loader component I would need to pass additional props and change some stylings

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
HIMDanny: unfortunately, we do. Without this div, the dots will squeeze. I can move it to a separate component if you want

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
liza-veis: Let's pass additional props to the existing loader, something like - isOverflow, and we can also wrap it in the div in loader component using conditional rendering

HIMDanny: sure. Then I will extend the component which we have and add this prop to existing ones

dmytro-kucherenko: Yeah, the same approach is needed in meditation modal while loading

HIMDanny: I updated all occurrences of Loader

v0ldemar01 (dismissed):

frontend/src/pages/chat/components/chat-layout/chat-layout.tsx
```diff
@@ -30,15 +31,21 @@ const ChatLayout: React.FC<Properties> = ({ filter }) => {
...
+  } = useAppSelector(({ chats, auth }) => {
+    return {
+      currentChatMessages: chats.currentChatMessages,
+      authenticatedUser: auth.authenticatedUser as UserAuthResponseDto,
+      createMessageDataStatus: chats.createMessageDataStatus,
+      generateReplyDataStatus: chats.generateReplyDataStatus,
+    };
+  });
   const hasId = Boolean(id);
+  const isChatbotReplyLoading = generateReplyDataStatus === 'pending';
```
fshabanov:
```suggestion
  const isChatbotReplyLoading = generateReplyDataStatus === DataStatus.PENDING;
```

fshabanov (changes requested):

OlenaYatseniuk (approved):

dmytro-kucherenko (approved):

v0ldemar01 (approved):

fshabanov (approved):

liza-veis (approved):

PR #470, opened 2023-09-28
cp-455: Improved Mobile Meditation Player Timestamps

iamAlinaaa:
I dont have Android device to check it. On my emulator all worked good (when choosing dark mode) even without applying these changes. But i have found this most rated approach to manually set these colors, and though it is look not that elegant, i hope it will solve the problem on devices, but we need to check it for sure.

Снимок экрана 2023-09-27 101650

### DIFF
--- a/mobile/src/libs/components/player/components/progress-bar/progress-bar.tsx
+++ b/mobile/src/libs/components/player/components/progress-bar/progress-bar.tsx
@@ -1,6 +1,5 @@
 import Slider from '@react-native-community/slider';
 import React from 'react';
-import { useColorScheme } from 'react-native';
 
 import { Text, View } from '#libs/components/components';
 import { AppColor } from '#libs/enums/enums';
@@ -8,30 +7,21 @@ import { getFormattedTime } from '#libs/helpers/helpers';
 import { usePlayerControls, usePlayerProgress } from '#libs/hooks/hooks';
 
 import { TRACK_START_TIME } from './libs/constants';
-import { darkStyles, styles } from './styles';
+import { styles } from './styles';
 
 type Properties = {
   isPlaying: boolean;
 };
 
 const ProgressBar: React.FC<Properties> = ({ isPlaying }) => {
-  const colorScheme = useColorScheme();
   const [timeProgress, duration] = usePlayerProgress();
 
   const { handleSeek } = usePlayerControls({ isPlaying });
 
   return (
     <View style={styles.wrapper}>
-      <Text
-        style={[styles.progress, colorScheme === 'dark' && darkStyles.progress]}
-      >
-        {getFormattedTime(timeProgress)}
-      </Text>
-      <Text
-        style={[styles.duration, colorScheme === 'dark' && darkStyles.duration]}
-      >
-        {getFormattedTime(duration)}
-      </Text>
+      <Text style={styles.progress}>{getFormattedTime(timeProgress)}</Text>
+      <Text style={styles.duration}>{getFormattedTime(duration)}</Text>
       <Slider
         minimumValue={TRACK_START_TIME}
         maximumValue={duration}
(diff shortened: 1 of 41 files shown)


mobile/src/libs/components/player/components/progress-bar/styles.ts
```diff
@@ -14,16 +14,32 @@ const styles = StyleSheet.create({
...
+});
+
+const darkStyles = StyleSheet.create({
+  progress: {
+    color: AppColor.GRAY_400,
+  },
+  duration: {
     color: AppColor.GRAY_400,
   },
 });
```
danilittaa: Why do we need two identical styles?

danilittaa (commented):
I'm not sure, but it seems to me that you've done unnecessary work

mobile/src/libs/components/player/components/progress-bar/styles.ts
iamAlinaaa: @danilittaa did you read that i have written in description ?)

danilittaa: Yes, did you read my comment? 
Why do we need two absolutely identical styles? (I would understand if you had used different colors, but you made everything the same)

iamAlinaaa: Arent we suppose to have the same colors ? If you think its not necessary please take android with already merged prs and check how does it work now with just one color setted.

danilittaa: For now, there's no point in creating two separate styles since the color for the light and dark themes is the same. You can simply reference the same file

iamAlinaaa: do you think it looks better and person who is not familiar with code will look at it and wont think that it not makes any lexical sence ?
Снимок экрана 2023-09-28 113025

mobile/src/libs/components/player/components/progress-bar/progress-bar.tsx
```diff
@@ -1,27 +1,34 @@
 import Slider from '@react-native-community/slider';
 import React from 'react';
+import { useColorScheme } from 'react-native';
```
danilittaa:
```suggestion
import { useColorScheme } from '#libs/hooks/hooks';
```

iamAlinaaa: we use it just for one time here

mobile/src/libs/components/player/components/progress-bar/styles.ts
danilittaa: I think we don't need this line at all. It seems to me (but I'm not sure) that the color issue in dark mode should have been resolved by simply adding color. That's why Lisa in the chat [here](https://bsa2023.slack.com/archives/C05P563E2MN/p1695832484447239?thread_ts=1695826530.921589&cid=C05P563E2MN) said that QA should check it first. Did you ask QA to verify it?

iamAlinaaa: I checked it in the morning form real device, there was this problem.

mobile/src/libs/components/player/components/progress-bar/progress-bar.tsx
dmytro-kucherenko: Nevertheless all hooks should be exported from their file, even once)

mobile/src/libs/components/player/components/progress-bar/progress-bar.tsx
```diff
@@ -1,27 +1,37 @@
...
   const [timeProgress, duration] = usePlayerProgress();
 
   const { handleSeek } = usePlayerControls({ isPlaying });
 
   return (
     <View style={styles.wrapper}>
-      <Text style={styles.progress}>{getFormattedTime(timeProgress)}</Text>
-      <Text style={styles.duration}>{getFormattedTime(duration)}</Text>
+      <Text
+        style={[styles.progress, colorScheme === 'dark' && darkStyles.progress]}
```
fshabanov: Can we ever have 'dark' here? You enforced the light mode in styles.xml file

fshabanov (changes requested):

mobile/src/libs/components/player/components/progress-bar/progress-bar.tsx
iamAlinaaa: @fshabanov okay, i can delete it, but as I said i dont have android device, so i couldnt check how this worked

liza-veis (approved):

PR #489, opened 2023-09-29
cp-427: Improvements for enpoints and swagger documentation

OlenaYatseniuk:
- changed status code for delete chat route to 200
- changed response for delete journal route to boolean
- delete unnecessary GET journal/id route
- added pattern for createJournalEntry validation schema to allow only strings with tags to prevent UI errors when making requests from Swagger or Postman


v0ldemar01 (commented):
@OlenaYatseniuk please check our postman config, ci postman tests fails of missing new changes there

HIMDanny (dismissed):

backend/src/packages/chats/chat.controller.ts
```diff
@@ -496,6 +496,13 @@ class ChatController extends BaseController {
     const { id } = options.params;
     const { id: userId } = options.user;
 
+    if (!Number(id)) {
```
v0ldemar01: Let's add it to service layer

backend/src/packages/journal-entries/journal-entry.controller.ts
```diff
@@ -396,6 +355,13 @@ class JournalEntryController extends BaseController {
       body: { title, text },
     } = options;
 
+    if (!Number(id)) {
```
v0ldemar01: Same

v0ldemar01 (changes requested):

shared/src/packages/journal/libs/enums/journal-entry-validation-message.enum.ts
```diff
@@ -3,6 +3,8 @@ import { JourneyEntryValidationRule } from './journal-entry-validation-rule.enum
 const JournalEntryValidationMessage = {
   TITLE_REQUIRED: 'Title is required.',
   TITLE_MUST_BE_LESS_THAN: `Title must be less than ${JourneyEntryValidationRule.MAXIMUM_TITLE_LENGTH} characters.`,
+  INVALID_TEXT_FORMAT:
+    'Invalid text format. It should start with "<p" and end with "</p>".',
```
v0ldemar01: Let's limit to Invalid text format, clients are not good at html tags

v0ldemar01 (changes requested):

shared/src/packages/journal/libs/constants/constants.ts
```diff
@@ -8,4 +8,8 @@ const NOTE_SANITIZER_OPTIONS: IOptions = {
   },
 };
 
-export { NOTE_SANITIZER_OPTIONS };
+const TEXT_PATTERN = /^<p\b[^>]*>.*<\/p>$/;
+
+const TEXT_PATTERN_NAME = 'textFormat';
```
liza-veis: text pattern should be in ValidationRule, and we don't need TEXT_PATTERN_NAME

liza-veis (changes requested):

backend/src/packages/chats/chat.service.ts
```diff
@@ -168,6 +168,13 @@ class ChatService implements Service {
     id: number;
     userId: number;
   }): Promise<boolean> {
+    if (!Number(id)) {
```
v0ldemar01:
```suggestion
    if (!id)) {
```
id is already number

backend/src/packages/journal-entries/journal-entry.service.ts
```diff
@@ -75,14 +80,23 @@ class JournalEntryService implements Service {
     title,
     text,
   }: JournalEntryUpdateRequestDto): Promise<JournalEntryGetAllItemResponseDto> {
+    if (!Number(id)) {
```
v0ldemar01:
```suggestion
    if (!id) {
```

v0ldemar01 (changes requested):

backend/src/packages/chats/chat.service.ts
OlenaYatseniuk: yes, thanks for noticing!

v0ldemar01 (approved):

PR #506, opened 2023-09-29
cp-342: Add tooltip to cards

OlesiaTr:
tooltip_2


frontend/src/libs/components/card/card.tsx
```diff
@@ -11,6 +11,7 @@ type Properties = {
   imageUrl?: string;
   onClick: () => void;
   isActive?: boolean;
+  isTooltipAvailable?: boolean;
```
OlenaYatseniuk, line 14:
```suggestion
  hasTooltip?: boolean;
```
Maybe it's better to rename it like this?

OlesiaTr: I wanted to write that, but I saw Liza was saying somewhere, that it's better if variables are written in one style

HIMDanny: @Liza-Veis said that we should stick to the one style in the app about naming, so maybe the name with the `is` prefix would suit better

OlenaYatseniuk: Okey, agree

OlenaYatseniuk (dismissed):

v0ldemar01 (dismissed):

frontend/src/libs/components/card/styles.module.scss
```diff
@@ -87,6 +87,44 @@
...
+  opacity: 0;
+  transition: opacity 0.2s ease;
+}
+
+.tooltip::after {
+  content: attr(data-tooltip);
+  position: absolute;
+  top: 110%;
+  left: 20%;
+  z-index: 300;
```
dmytro-kucherenko, line 108: It's no reason to make it so high, 10 will be more than enough

HIMDanny (approved):

PR #507, opened 2023-09-29
cp-492: Fix User Can't Play The Updated Meditation

danilittaa:
In the ticket @k-mlch described the issue with meditations, which was because mocked data was still being used at that time. But now I have encountered a different problem - when I play a track, a completely different one starts playing (only the first one)

The issue was that by default, `react-native-track-player` sets the first track added to the playlist as the initial track. So, I had to manually set the inital track to make everything work correctly

### DIFF
--- a/mobile/src/screens/meditation/meditation.tsx
+++ b/mobile/src/screens/meditation/meditation.tsx
@@ -31,7 +31,7 @@ const Meditation: React.FC = () => {
       const currentTrackIndex = meditationEntries.findIndex((item) => {
         return item.id === selectedMeditationEntry.id;
       });
-      void player.setFirstTrack(Number(currentTrackIndex));
+      void player.setFirstTrack(currentTrackIndex);
     }
   }, [selectedMeditationEntry, meditationEntries]);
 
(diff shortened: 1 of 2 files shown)


mobile/src/screens/meditation/meditation.tsx
```diff
@@ -13,15 +14,27 @@ type RouteParameters = {
...
 
   const route = useAppRoute();
   const { duration }: RouteParameters = route.params as RouteParameters;
 
+  useEffect(() => {
+    if (selectedMeditationEntry) {
+      const currentTrackIndex = meditationEntries.findIndex((item) => {
+        return item.id === selectedMeditationEntry.id;
+      });
+      void player.setFirstTrack(Number(currentTrackIndex));
```
liza-veis: we don't need number conversion, but we need to handle case where track is not found
```suggestion
      void player.setFirstTrack(currentTrackIndex);
```

liza-veis (changes requested):

mobile/src/screens/meditation/meditation.tsx
danilittaa: We already have `if ` to check if selectedMeditationEntry exists. If it doesn't, the first track in the queue will be set as the default first track

v0ldemar01 (approved):

fshabanov (approved):

liza-veis (approved):

PR #510, opened 2023-09-30
cp-509: Add New Note Button Doesn't Work

danilittaa:
issue: " I am getting an error that says 'title is required,' and instead of an empty note, some other content is displayed. I thinks this is happening because, for some reason, adding notes has slightly different logic than creating a chat or creating a note on the frontend. Here, we seem to create a new note immediately, even when the user hasn't entered anything. Perhaps the validation for the title was added later than when Alina finished her task, so she didn't see this error"

- moved the logic of creating a note from the moment the user presses the button to the moment the user enters something

https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/108797614/462956f0-71d4-404c-a738-e47944cd9865

### DIFF
--- a/mobile/src/screens/note/note.tsx
+++ b/mobile/src/screens/note/note.tsx
@@ -65,13 +65,22 @@ const Note: React.FC = () => {
 
   const handleSaveNote = useCallback(
     (data: JournalEntryCreateRequestDto) => {
-      void dispatch(
-        journalActions.updateJournalEntry({
-          id: Number(id),
-          title: data.title || EMPTY_NOTE_TITLE,
-          text: data.text,
-        }),
-      );
+      if (id) {
+        void dispatch(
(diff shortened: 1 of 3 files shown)


OlenaYatseniuk (dismissed):

mobile/src/screens/note/note.tsx
```diff
@@ -81,7 +90,7 @@ const Note: React.FC = () => {
   }, [handleSaveNote]);
 
   useEffect(() => {
-    if (id && isDirty) {
+    if (isDirty) {
```
liza-veis, line 93: shouldn't we also remove id from dependencies?

liza-veis (changes requested):

v0ldemar01 (approved):

liza-veis (approved):

PR #512, opened 2023-09-30
cp-503: Mobile Delete Note And Card Improvement

iamAlinaaa:
After user slides for deleting an option, once he/she clicks somewhere else on the screen, the option becomes not in 'delete' state and when another card is swiped previos one is swiped back so delete option is available only for one card at a time

https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/111873795/6d03b46a-e57c-46ae-a458-2778f7c638e4


mobile/src/libs/components/card/card.tsx
```diff
@@ -34,7 +39,28 @@ const Card: React.FC<Properties> = ({
   onPress,
   iconRight,
   onIconPress,
+  id,
+  isModalVisible,
 }) => {
+  const handleSwipeableReference = useCallback(
```
v0ldemar01, line 45: Can we prefer using useRef with array without Map outside?

v0ldemar01 (changes requested):

OlesiaTr (approved):

PR #513, opened 2023-09-30
cp-434: Added Open Ai errors

Nanaier:
Screenshot 2023-09-30 at 12 36 05

### DIFF
--- a/backend/src/libs/packages/api/base-http-api.ts
+++ b/backend/src/libs/packages/api/base-http-api.ts
@@ -6,7 +6,7 @@ import {
   HTTPHeader,
 } from '#libs/packages/http/http.js';
 import { HTTPError } from '#libs/packages/http/http.js';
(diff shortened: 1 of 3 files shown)


backend/src/libs/packages/api/base-http-api.ts
```diff
@@ -83,17 +83,11 @@ class BaseHttpApi implements HTTPApi {
...
-
-    try {
-      parsedException = (await response.json()) as ServerErrorResponse;
-    } catch {
-      parsedException = {
-        errorType: ServerErrorType.COMMON,
-        message: response.statusText,
-      };
-    }
+  private handleError(response: Response): Promise<never> {
```
liza-veis, line 86: we also need to catch and rethrow error in service, because we need to override status code, for example if open ai respond with 401, we don't want to respond with 401 because user will be signed out

liza-veis (changes requested):

dmytro-kucherenko (dismissed):

PR #516, opened 2023-09-30
cp-363: Scroll to selected chat after reloading

dmytro-kucherenko:
Done: 
- the chat list state displays the list with the previously selected chat after reloading the page when a chat that doesn't fit on the page is selected

https://github.com/BinaryStudioAcademy/bsa-2023-calmpal/assets/92298163/55b380dd-2e35-42c0-ba74-53c1f6e051f9


frontend/src/pages/chat/components/chat-sidebar/chat-sidebar.tsx
```diff
@@ -57,6 +60,16 @@ const ChatSidebar: React.FC<Properties> = ({
     void dispatch(chatsActions.getAllChats(filter));
   }, [dispatch, filter]);
 
+  useEffect(() => {
+    if (chatsDataStatus === DataStatus.FULFILLED) {
+      setIsChatsLoaded(true);
+    }
+  }, [chatsDataStatus, setIsChatsLoaded]);
```
HIMDanny, line 67: why do we need a state for this? Can't we simply check like this?
```ts
const isChatsLoaded = chatsDataStatus === DataStatus.FULFILLED
```

dmytro-kucherenko: @HIMDanny We can't. Scroll to the selected card should be only on first chats loading, because then data status will change on every search request

HIMDanny: now I see, thanks for clarification

HIMDanny: btw, we also can use ref for this to not cause a rerender after `setIsChatLoaded(true)`

dmytro-kucherenko: @HIMDanny What do you mean by saying use ref?

HIMDanny: instead of using state for setting this boolean value, we can make something like:
```ts
const isChatLoaded = useRef(false);
...
isChatLoaded.current = true;
```

HIMDanny: nevermind, now I am not sure whether it will work)

